PR #3, opened 2019-07-01
Update Model-class.R

TJaki:
Altered dose standardisation in DualEndpointEmax to fix bug when trying to simulate


danielinteractive:
Thanks Thomas,
could you please also check if the same change should be applied to the other DualEndpoint... classes?

TJaki:
Checked and made similar change to DualEndpointBeta (others were ok).

PR #13, opened 2021-11-12
Add Github actions, remove docs

dinakar29:
Adds common Github actions and removes docs from the main branch. Docs will be published on the gh-pages branch instead.


danielinteractive (approved):
thx Dinakar!

PR #14, opened 2021-11-16
Added lifcycle related figures + modify man of  Validate()

wwojciech:
Added lifcycle related figures with usethis::use_lifecycle() + modify man of Validate().


R/helpers.R
```diff
@@ -358,6 +358,9 @@ printVignette <- function (x, ...)
 
 
 ##' A Reference Class to help programming validation for new S4 classes
+##' 
+##'
+##' @description `r lifecycle::badge("experimental")`
```
danielinteractive, line 363: as mentioned please use 
```suggestion
##' @description `r lifecycle::badge("stable")`
```

DESCRIPTION
```diff
@@ -25,6 +25,7 @@ Depends:
     graphics,
     R (>= 3.0.0)
 Imports:
+    lifecycle,
```
danielinteractive, line 28: I think because of this we also need to import a single function from this package. please add a corresponding roxygen line to https://github.com/Roche/crmPack/blob/main/R/crmPack-package.R
(otherwise R CMD check will complain)

danielinteractive: see e.g. https://github.com/insightsengineering/hermes/blob/5ad69c62b0b7dec08e23829dffbcb4d8be1d5ca6/R/package.R#L17 for comparison

danielinteractive (approved):
thx Wojciech, please see 2 comments inside, good to merge after addressing those

DESCRIPTION
wwojciech: Sure, thank you for spotting this!

R/helpers.R
wwojciech: Done, thanks!

PR #15, opened 2021-11-17
Set up the package to use testthat + Rproj file

wwojciech:
Set up the package to use testthat.
Add Rproj file


danielinteractive (approved):
thx Wojciech!

PR #16, opened 2021-11-18
Clean-up Validate class 

### LINKED ISSUE #10: Clean-up `Validate` class
To do:

- [x] Revise documentation
  - [x] RStudio outline section dividers
  - [x] Replace `##'` by `#'`
  - [x] Add to `_pkgdown.yaml`
- [x] Add unit tests
  - [x] for classes: 
  - [x] for functions: as usual
- [x] Refactor as needed
  - [x] ...

wwojciech:
closes #10 

Clean-up Validate class: 
1. Revise documentation
 - RStudio outline section dividers
 - Add lifecycle badge (as a default take "stable" for existing functions or classes)
- Replace ##' by #'
2. Add to _pkgdown.yaml
 3. Add unit tests
 - Improve code styling (indentation etc.) with styler


_pkgdown.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - title: Methods
+    contents:
+  - title: Functions
+    contents:
+  - title: Graphs
+    contents:
+  - title: Control Functions
+    contents:
+  - title: Assertions
+    contents:
```
danielinteractive, line 44:
```suggestion
    contents:

```
(newline at eof)

R/helpers.R
```diff
@@ -356,36 +356,41 @@ printVignette <- function (x, ...)
...
+#' @field msg the message character vector
+Validate <- setRefClass(
+  Class = "Validate",
+  fields = list(msg = "character"),
+  methods = list(
+    check = function(test, string) {
+      if (test) {
+        NULL
+      } else {
+        msg <<- c(msg, string)
```
danielinteractive, line 382: since we use here the `<<-` assignment operator, can you check in a dedicated test that having a `msg` variable in the test does not confuse this method?

R/helpers.R
```diff
@@ -356,36 +356,41 @@ printVignette <- function (x, ...)
...
+#' that is returning `FALSE` the vector gets a new element - the string
+#' explaining the failure of the validation.
+#'
+#' @name Validate
+#' @field msg the message character vector
+Validate <- setRefClass(
+  Class = "Validate",
+  fields = list(msg = "character"),
+  methods = list(
+    check = function(test, string) {
```
danielinteractive, line 378: how can we best document the `check` and `result` methods of this class with `roxygen2`? maybe there are some examples in other packages? if there is no dedicated functionality in `roxygen2` maybe just in `details` or so? similar like with `R6` documentation?

danielinteractive (commented):
thx Wojciech, nice, pls see comments inside

R/helpers.R
wwojciech: Thank you @danielinteractive , I think so called `docstring` (see https://r-pkgs.org/man.html) might be appropriate for this case. I will create this entry as a proposal.

R/helpers.R
wwojciech: @danielinteractive - Do you mean dedicate unit test? If it so, sure, I will create something I believe is ok - but please double check it after I commit it, since I am not so sure if it will be ok. Also I think there should not be any confusions since, the msg ("global") variable stays enclosed in the environment of the instance of `Variable` class.

danielinteractive: yeah dedicated test. agree it should be ok but good to be sure.

wwojciech: Absolutely, I added 4 quite different tests - please decide which one of them should stay and which should be removed. I hope there are differ enough so that they can stay all :):).

R/helpers.R
```diff
@@ -356,36 +356,41 @@ printVignette <- function (x, ...)
...
+      if (test) {
```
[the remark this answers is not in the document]
wwojciech: Would you prefer to import just specific methods (`assert_flag`, `assert_string`) as for now, or the whole `checkmate` as it is in `helios`?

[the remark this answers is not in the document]
danielinteractive: whole package

danielinteractive (approved):

inst/WORDLIST
```diff
@@ -316,6 +317,7 @@ STUDYID
 subclasses
 subfolder
 subtractive
+succesull
```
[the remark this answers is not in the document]
wwojciech: Sure, thank you.

PR #25, opened 2021-11-23
20 Data clean up

### LINKED ISSUE #20: `Data` clean up
To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] for classes:
  - [x] for functions: as usual
- [x] Refactor as needed

wwojciech:
closes #20


R/Data-class.R
```diff
@@ -30,147 +30,117 @@ NULL
...
+#' i.e. the dose grid.
+#' @slot nGrid (`integer`)\cr number of gridpoints.
+#' @slot xLevel (`integer`)\cr the levels for the doses the patients have been given,
+#' w.r.t `doseGrid`.
+#' @slot placebo (`logical`)\cr if `TRUE` the first dose level
+#' in the `doseGrid`is considered as PLACEBO.
+#'
+#' @example examples/Data-class.R
+#' @export
+#' @keywords classes
```
danielinteractive, line 54: I think we can remove these `@keywords` tags in general. since in the `pkgdown` information we already group things nicely - this is duplicate grouping

R/Data-class.R
```diff
@@ -30,147 +30,117 @@ NULL
...
+    ID = safeInteger(ID),
+    cohort = safeInteger(cohort),
+    doseGrid = doseGrid,
+    nObs = length(x),
+    nGrid = length(doseGrid),
+    xLevel = matchTolerance(x = x, table = doseGrid),
+    placebo = placebo
+  )
 
+  return(ret)
```
danielinteractive, line 142: avoid explicit return statements, just let the function return implicitly the last value in the function body

R/Data-validity.R
```diff
@@ -1,35 +1,120 @@
-#' Internal Helper Functions for Validation of `GeneralData` Objects.
+#' Internal Helper Functions for Validation of `GeneralData` and `Data` Objects.
```
danielinteractive, line 1: I would not try to list all classes here. otherwise title becomes too long with the other 5 classes.

R/Data-validity.R
```diff
@@ -1,35 +1,120 @@
...
-#' [`GeneralData-class`] object and therefore not exported.
+#' [`GeneralData-class`] and [`Data-class`] object and therefore not exported.
 #'
 #' @name validate_data
 #' @param object (`GeneralData`)\cr object to validate.
 #' @return A character vector with the validation failure messages,
 #' or `TRUE` in case validation passes.
 NULL
 
+#' @describeIn validate_data helper function.
```
danielinteractive, line 14: move this one function down such that it is on top of `validate_data`

R/Data-validity.R
```diff
@@ -1,35 +1,120 @@
...
   o$check(!is.unsorted(object@cohort, strictly = FALSE), "cohort indices must be sorted")
-  
+
+  o$result()
+}
+
+#' @describeIn validate_data validates that the [`Data-class`] object contains
+#' valid elements with respect to their types, dependency and length.
+#'
+validate_data <- function(object) {
```
danielinteractive, line 51: this function is too long (> 50 lines as rule of thumb) - let's think about logical parts we could break this into

tests/testthat/test-Data-validity.R
```diff
@@ -6,5 +6,89 @@ test_that("validate_subjects returns TRUE for valid object", {
 })
 
 test_that("validate_subjects returns error message for non-valid object", {
-  expect_error(.GeneralData(ID = 1:3, cohort = 3:5, nObs = 4))
+  expect_error(.GeneralData(ID = 1:3, cohort = 3:5, nObs = 4), regexp = ".+")
+})
+
+# validate_data ----
+
+test_that("validate_data returns TRUE for valid object", {
```
danielinteractive, line 14: this should directly call `validate_data` instead of `.Data`

danielinteractive (commented):
thx Wojciech, pls see first comments inside, talk to you later too

tests/testthat/test-Data-class.R
```diff
@@ -1,7 +1,31 @@
...
+  expect_silent(.GeneralData(ID = 1:3, cohort = 3:5, nObs = 3L))
+})
+
+test_that(".GeneralData throws error for non-valid object", {
+  expect_error(.GeneralData(ID = 1:3, cohort = 3:5, nObs = 4), regexp = ".*")
+})
+
 # Data ----
 
 test_that("Data can be created with default constructor Data.", {
```
danielinteractive, line 13:
```suggestion
test_that("Data can be created with default constructor .Data", {
```

tests/testthat/test-Data-class.R
```diff
@@ -1,7 +1,31 @@
...
   res <- expect_silent(Data())
```
danielinteractive, line 14:
```suggestion
  res <- expect_silent(.Data())
```

tests/testthat/test-Data-class.R
```diff
@@ -1,7 +1,31 @@
...
   expect_s4_class(res, "Data")
   expect_true(validObject(res))
 })
+
```
danielinteractive, line 18:
```suggestion
# Data-constructor ----
```

tests/testthat/test-Data-class.R
```diff
@@ -1,7 +1,31 @@
...
+test_that(".GeneralData executes without error for valid object", {
+  expect_silent(.GeneralData(ID = 1:3, cohort = 3:5, nObs = 3L))
+})
+
+test_that(".GeneralData throws error for non-valid object", {
+  expect_error(.GeneralData(ID = 1:3, cohort = 3:5, nObs = 4), regexp = ".*")
+})
+
 # Data ----
```
danielinteractive, line 11:
```suggestion
# Data-class ----
```

tests/testthat/test-Data-class.R
```diff
@@ -1,7 +1,31 @@
...
 
 test_that("Data can be created with default constructor Data.", {
   res <- expect_silent(Data())
   expect_s4_class(res, "Data")
   expect_true(validObject(res))
 })
+
+test_that("Data can be created for valid object", {
```
danielinteractive, line 19: make sure separate tests for Data() cover the code

tests/testthat/test-Data-validity.R
```diff
@@ -6,5 +6,90 @@ test_that("validate_subjects returns TRUE for valid object", {
 })
 
 test_that("validate_subjects returns error message for non-valid object", {
-  expect_error(.GeneralData(ID = 1:3, cohort = 3:5, nObs = 4))
+  expect_error(.GeneralData(ID = 1:3, cohort = 3:5, nObs = 4), regexp = ".+")
```
danielinteractive, line 9: let's do like in helios - construct input object, make one single thing wrong/invalid, then run validate function on it, and expect specific error message

tests/testthat/test-Data-validity.R
```diff
@@ -6,5 +6,90 @@ test_that("validate_subjects returns TRUE for valid object", {
...
+  )
+})
+
+test_that("validate_data returns error message for non-valid object
+          (multiplcbe doses in cohort 1)", {
+  plcb <- 0.001
+  x <- c(plcb, 25, 25, 300, plcb, 50, 50, 50, plcb, 100, 100, 100)
+  dose_grid <- c(plcb, seq(25, 300, 25))
+  expect_error(
+    .Data(
```
danielinteractive, line 61: can put like a small helper function in `helper-data.R` in `testthat` directory - e.g. `h_get_data()` which just produces this same good Data thing

R/Data-validity.R
```diff
@@ -17,19 +17,104 @@ NULL
...
+  x_eq_grid_xLev <- all.equal(
+    object@x,
+    object@doseGrid[object@xLevel],
+    tolerance = 1e-10,
+    check.names = FALSE,
+    check.attributes = FALSE
+  )
+
+  o$check(
+    test_flag(x_eq_grid_xLev) && x_eq_grid_xLev == TRUE,
```
danielinteractive, line 101: maybe put this in another helper?

wwojciech:
@danielinteractive Daniel, do you really want to look in unit tests for specific error message? Sometimes the error messages are quite involved and contain special characters " " , for example:
```
test_that(".GeneralData throws error for non-valid object", {
  expect_error(
    .GeneralData(ID = 1:3, cohort = 3:5, nObs = 4),
    regexp = "Error: invalid class \"GeneralData\" object: invalid object for slot \"nObs\" in class \"GeneralData\": got class \"numeric\", should be or extend class \"integer\""
  )
})
```
and this still fails for some reason

NAMESPACE
```diff
@@ -221,6 +221,7 @@ exportMethods(update)
 import(checkmate)
 importFrom(GenSA,GenSA)
 importFrom(MASS,ginv)
+importFrom(checkmate,test_flag)
```
danielinteractive: not necessary - as we import the whole package (see line 221)

R/Data-validity.R
```diff
@@ -89,16 +90,8 @@ validate_data <- function(object) {
...
-    object@x,
-    object@doseGrid[object@xLevel],
-    tolerance = 1e-10,
-    check.names = FALSE,
-    check.attributes = FALSE
-  )
-
   o$check(
-    test_flag(x_eq_grid_xLev) && x_eq_grid_xLev == TRUE,
+    all_equal(object@x, object@doseGrid[object@xLevel]),
```
danielinteractive, line 94: nice ! thx

R/helpers.R
```diff
@@ -510,3 +510,27 @@ myBarplot <- function(x, description, xaxisround=0)
                          round(dat$x, xaxisround))
 }
 
+#' Wrapper for all.equal
```
danielinteractive, line 513: hm ok I see - can you double check if `expect_equivalent` does not already do something very similar maybe? now that I see this I remember that function

examples/Data-class.R
```diff
@@ -1,4 +1,5 @@
 # create some data from the class 'Data'
-myData <- Data(x = c(0.1, 0.5, 1.5, 3, 6, 10, 10, 10),
-               y = c(0, 0, 0, 0, 0, 0, 1, 0),
-               doseGrid = c(0.1, 0.5, 1.5, 3, 6, seq(from = 10, to = 80, by = 2)))
\ No newline at end of file
+my_data <- Data(x = c(0.1, 0.5, 1.5, 3, 6, 10, 10, 10),
+                y = c(0, 0, 0, 0, 0, 0, 1, 0),
+                doseGrid = c(0.1, 0.5, 1.5, 3, 6,
+                             seq(from = 10, to = 80, by = 2)))
```
danielinteractive, line 5:
```suggestion
my_data <- Data(
  x = c(0.1, 0.5, 1.5, 3, 6, 10, 10, 10),
  y = c(0, 0, 0, 0, 0, 0, 1, 0),
  doseGrid = c(
    0.1, 0.5, 1.5, 3, 6,
    seq(from = 10, to = 80, by = 2)
  )
)
```
-> code formatting as usual for long function calls please

tests/testthat/helper-data.R
```diff
@@ -0,0 +1,26 @@
...
+#' for an example data.
+#'
+#' @return The [`Data-class`] object.
+#'
+h_get_data <- function() {
+  plcb <- 0.001
+  x <- c(plcb, 25, 25, 25, plcb, 50, 50, 50, plcb, 100, 100, 100)
+  dose_grid <- c(plcb, seq(25, 300, 25))
+
+  data <- .Data(
```
danielinteractive, line 15:
```suggestion
  .Data(
```
avoid assigning to variable when it is the last value in function body

tests/testthat/test-Data-class.R
```diff
@@ -1,31 +1,83 @@
-# GeneralData ----
+# GeneralData-class ----
 
-test_that(".GeneralData executes without error for valid object", {
+test_that("GeneralData default constructor .GeneralData
+          executes without error for valid object", {
```
danielinteractive, line 4: if you break the title string please use `paste()` otherwise we have all this whitespace and new line in the string.

tests/testthat/test-Data-class.R
```diff
@@ -1,31 +1,83 @@
...
+        x = c(plcb, 25, 25, 25, plcb, 50, 50, 50, plcb, 100, 100, 100),
+        y = c(0L, 0L, 0L, 0L, 0L, 0L, 0L, 0L, 0L, 0L, 1L, 0L),
+        doseGrid = c(plcb, seq(25, 300, 25)),
+        placebo = FALSE
+      ),
+      regexp = "Used default patient IDs!"
+    ),
+    regexp = "Used best guess cohort indices!"
+  )
+})
```
danielinteractive, line 83: not necessary but ok.

tests/testthat/test-Data-validity.R
```diff
@@ -6,90 +6,62 @@ test_that("validate_subjects returns TRUE for valid object", {
 })
 
 test_that("validate_subjects returns error message for non-valid object", {
-  expect_error(.GeneralData(ID = 1:3, cohort = 3:5, nObs = 4), regexp = ".+")
+  object <- expect_silent(.GeneralData(ID = 1:3, cohort = 3:5, nObs = 3L))
+  object@nObs <- 4L
+  expect_equal(
+    object = validate_subjects(object),
+    expected = c(
```
danielinteractive, line 13: generally for these expect functions - no need to give explicit argument names. can just say `expect_*(result, expected)` without `object =` etc.

danielinteractive (commented):
thx Wojciech, few more comments

NAMESPACE
wwojciech: I have just spotted it, thanks!

R/helpers.R
wwojciech: `expect_equivalent ` seems to be deprecated by `expect_equal`. And `expect_equal` suffers from the same problems as `all.equal` which we solved using by the wrapper `all_equal`.

examples/Data-class.R
wwojciech: I did it intentionally as this is an example printed in the function man page, and in this case seems that non-code formatting is used for easy reading. See, e.g. `?read.table`. Would you like me to still change it?

danielinteractive: yeah I changed it now

tests/testthat/helper-data.R
```diff
@@ -0,0 +1,26 @@
+#' `h_get_data`
+#'
+#' @description `r lifecycle::badge("stable")`
+#'
+#' This is the helper function which creates a [`Data-class`] object
+#' for an example data.
+#'
+#' @return The [`Data-class`] object.
+#'
```
danielinteractive, line 9: No need to document these helper functions

danielinteractive (commented):
thx Wojciech - I did some additional work now on this branch. please have a look at my commit to see what changed and let me know if any questions.
Can you please add unit tests for the two helper functions `h_doses_unique_per_cohort` and `h_all_equivalent`? Afterwards it should be good to merge. thx!

R/Data-validity.R
```diff
@@ -1,114 +1,100 @@
...
+      test_set_equal(object@cohort, object@cohort[!is_placebo]),
       "A cohort with only placebo is not allowed"
     )
-  } else {
-    plcb <- FALSE
+    o$check(
+      h_doses_unique_per_cohort(dose = object@x[!is_placebo], cohort = object@cohort[!is_placebo]),
+      "There must be only one dose level, other than placebo, per cohort"
+    )
   }
```
wwojciech, line 98: @danielinteractive I think this above check should also be conduced when `object@placebo` is `TRUE`, that is why I placed it out of `if-else` clause.

R/Data-validity.R
```diff
@@ -1,114 +1,100 @@
...
-
-  o$check(
-    h_is_dose_unique(dose = object@x[!plcb], cohort = object@cohort[!plcb]),
-    "There must be only one dose level, other than placebo, per cohort"
-  )
-
   o$result()
```
wwojciech, line 99: @danielinteractive I think this check should also be conducted when `object@placebo` is `FALSE`.

danielinteractive: @wwojciech ah ok - but then it should not use the `plcb` variable as it does not make sense then

wwojciech:
Thank you @danielinteractive I added unit tests as you asked and also polished few other things. I would also check with you on the following:

I have found that `base::all.equal` is a bit more sophisticated than I thought, I mean it by default compares on relative scale. So for example 
```
> all.equal(
+     target = 5,
+     current = 5.2,
+     tolerance = 0.1,
+     check.names = FALSE,
+     check.attributes = FALSE,
+ )
[1] TRUE
> all.equal(
+     target = 1,
+     current = 1.2,
+     tolerance = 0.1,
+     check.names = FALSE,
+     check.attributes = FALSE,
+ )
[1] "Mean relative difference: 0.2"
> 
```  
As you can see it is quite awkward, I mean the interpretation of `tolerance`, maybe we should always compare on absolute scale, i.e.:
```
> all.equal(
+     target = 5,
+     current = 5.2,
+     tolerance = 0.1,
+     check.names = FALSE,
+     check.attributes = FALSE,
+     scale = 1
+ )
[1] "Mean absolute difference: 0.2"
> 
```
?

danielinteractive:
@wwojciech relative difference is what makes sense in numeric precision comparisons, and users are used to it - so I would keep the default behavior.

R/helpers.R
```diff
@@ -520,14 +520,18 @@ myBarplot <- function(x, description, xaxisround=0)
...
 #' @return `TRUE` when `target` and `current` do not differ
 #'   up to desired tolerance and without looking at names or other attributes,
 #'   `FALSE` otherwise.
 #'
 h_all_equivalent <- function(target, 
                              current, 
                              tolerance = 1e-10) {
+  assert_numeric(target)
+  assert_numeric(current)
+  assert_number(tolerance)
```
danielinteractive, line 533: note that I did not assert this argument before because it can be done by `all.equal` but it is ok to do it.

danielinteractive (approved):
thx Wojciech!

PR #26, opened 2021-11-24
Use reusable workflows

dinakar29:
Use centralized Github Action workflows.


danielinteractive (approved):
thx @dinakar29 ! looks good. there is some failing latex problem that we did not have before though - maybe something that can be fixed?

dinakar29:
> thx @dinakar29 ! looks good. there is some failing latex problem that we did not have before though - maybe something that can be fixed?

Yes, fixing this now.

danielinteractive:
@dinakar29 not sure what is happening here, but it is strange that only with the workflow update the latex fails right? maybe you can check what is the difference re: latex handling going from individual to reusable workflow? I would prefer not to mess with the contents themselves (as it worked before)

dinakar29:
> @dinakar29 not sure what is happening here, but it is strange that only with the workflow update the latex fails right? maybe you can check what is the difference re: latex handling going from individual to reusable workflow? I would prefer not to mess with the contents themselves (as it worked before)

Yeah I'm looking into the differences right now. The JSSstyle LaTeX package might be out of date with the newer version of TeX, which could explain the underlying issue. I am able to consistently reproduce the issue. I'll provide more details soon as I determine the root cause. 

As a side note - are you intending to keep the old JSS submissions in this repo? There are tarballs, PDFs and LaTeX compile logs and artifacts, which, frankly should not be stored in version control.

danielinteractive:
@dinakar29 you are right - thanks for reminding me of the old baggage. I now pushed a commit here where I deleted a  lot of the old JSS folders and files. maybe that helps?

dinakar29:
> @dinakar29 you are right - thanks for reminding me of the old baggage. I now pushed a commit here where I deleted a lot of the old JSS folders and files. maybe that helps?

Thanks! I'll sort this out. I'm working on identifying and fixing the root cause and will resolve it soon.

crmPack.Rproj
```diff
@@ -9,7 +9,7 @@ UseSpacesForTab: Yes
 NumSpacesForTab: 2
 Encoding: UTF-8
 
-RnwWeave: Sweave
+RnwWeave: knitr
```
dinakar29, line 12: This makes it convenient to compile the PDF within RStudio

vignettes/crmPack-jss-paper.Rnw
```diff
@@ -27,6 +24,8 @@
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 %% declarations for jss.cls %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+\definecolor{Red}{rgb}{1.0, 0.0, 0.0}
+\definecolor{Blue}{rgb}{0.0, 0.0, 1.0}
```
dinakar29, line 28: This should resolve the color definitions issue.

PR #29, opened 2021-11-26
Closes #23

### LINKED ISSUE #23: `DataMixture` clean up
To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] for classes:
  - [x] for functions: as usual
- [x] Refactor as needed

wwojciech:
Closes #23


R/Data-class.R
```diff
@@ -178,151 +179,126 @@ DataDual <- function(w = numeric(),
...
+#'   for the share patients.
+#' @param ... parameters passed to [Data()].
+#'
+#' @export
+#' @example examples/Data-class-DataMixture.R
+DataMixture <- function(xshare = numeric(),
+                        yshare = integer(),
+                        ...) {
+  assert_numeric(xshare)
+  assert_integer(yshare, lower = 0, upper = 1)
```
danielinteractive, line 292:
```suggestion
```

tests/testthat/helper-data.R
```diff
@@ -25,3 +25,33 @@ h_get_data_dual <- function() {
...
+    doseGrid = d@doseGrid,
+    placebo = d@placebo,
+    ID = d@ID,
+    cohort = d@cohort
+  )
+}
+
+h_get_data_mixture <- function() {
+  d <- h_get_data()
+  DataMixture(
```
danielinteractive, line 46: can use `.DataMixture(d, ...)`

R/Data-validity.R
```diff
@@ -113,3 +113,52 @@ validate_data_dual <- function(object) {
...
+
+#' @describeIn validate_data_objects validates that the [`DataMixture`] object
+#' contains valid elements with respect to their types, dependency and length.
+validate_data_mixture <- function(object) {
+  o <- Validate()
+  # This below check is a bit obsolete as below invocations to
+  # test_* throw an error when len parameter (here initiated to object@nObsshare)
+  # is not a scalar.
+  o$check(
+    test_int(object@nObsshare),
```
danielinteractive, line 148: let's use if else and return early if not ok

R/Data-validity.R
```diff
@@ -16,10 +16,10 @@ NULL
 #'   `ID` and `cohort` vectors which are of the same length `nObs`.
 validate_subjects <- function(object) {
   o <- Validate()
-  o$check(
-    test_int(object@nObs),
-    "nObs must be of type integer of length 1"
-  )
+  if (!test_int(object@nObs)) { # in if clause so that below test_* won't fail
```
danielinteractive, line 19: longer comments better in separate line above, and please use sentence style (so start with upper case letter, end with full stop). thx

R/Data-validity.R
```diff
@@ -16,10 +16,10 @@ NULL
...
+    o$check(FALSE, "nObs must be of type integer of length 1")
+    return(o$result())
```
danielinteractive, line 21: here I would directly put the simpler
```suggestion
    return("nObs must be of type integer of length 1")
```

R/Data-validity.R
```diff
@@ -136,3 +136,27 @@ validate_data_parts <- function(object) {
...
+validate_data_mixture <- function(object) {
+  o <- Validate()
+
+  o$check(
+    test_subset(object@xshare, object@doseGrid),
+    "Dose values in xshare must be from doseGrid"
+  )
+  if (!test_int(object@nObsshare)) { # in if clause so that below test_* won't fail
+    o$check(FALSE, "nObsshare must be of type integer of length 1")
+    return(o$result())
```
danielinteractive, line 151: same comment here

tests/testthat/test-Data-validity.R
```diff
@@ -157,3 +157,62 @@ test_that("validate_data_parts returns error for part1Ladder with wrong values",
...
+
+  # We assign nObsshare of length different than 1,
+  object@nObsshare <- 1:5
+
+  # validate_data_mixture will throw the error here, which will be originated by
+  # test_* functions (used in validate_data_mixture) when the len parameter
+  # (initiated to object@nObsshare) is not a scalar.
+  expect_error(
+    validate_data_mixture(object),
+    regexp = "^Argument '[[:alnum:]]+' must have length 1$"
```
danielinteractive, line 216: this should no longer be the case now?

danielinteractive (commented):
thx - please see inside some simplification/formatting and question on test

R/Data-validity.R
wwojciech: Sure, but that would assume that there will be no tests defined before this check.....

danielinteractive (approved):

PR #32, opened 2021-11-29
Closes #24

### LINKED ISSUE #24: `DataDA` clean up
To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] for classes:
  - [x] for functions: as usual
- [x] Refactor as needed

wwojciech:
Closes #24


R/Data-validity.R
```diff
@@ -160,3 +160,31 @@ validate_data_mixture <- function(object) {
...
+  )
+  o$check(
+    test_numeric(object@Tmax, lower = 0 + .Machine$double.xmin, any.missing = FALSE),
+    "DLT window needs to be of type double and greater than 0"
+  )
+  # # DSB: why check not required?
+  # o$check(
+  #   test_numeric(object@t0, lower = 0, len = object@nObs, any.missing = FALSE),
+  #   "t0 entries must be non-negative of nObs length"
+  # )
```
danielinteractive, line 188: either remove this or even better include it - just see if examples/vignettes are compiled ok (done in automation checks)

tests/testthat/helper-data.R
```diff
@@ -40,3 +40,13 @@ h_get_data_mixture <- function() {
     nObsshare = 4L
   )
 }
+
+h_get_data_DA <- function() { # nolintr
```
danielinteractive, line 44:
```suggestion
h_get_data_augmented <- function() {
```
and similarly in the other test file where it is used

R/Data-class.R
```diff
@@ -290,95 +290,61 @@ DataMixture <- function(xshare = numeric(),
...
+#'
+#' [`DataDA`] is a class for the time-to-DLT augmented data.
+#' It inherits from [`Data`] and it contains additional DLT free survival times.
+#'
+#' @note `survival time` here refers to the time period for which the subject
+#' did not experience any DLT, and is not referring to deaths.
+#'
+#' @slot u (`numeric`)\cr the continuous vector of DLT free survival times.
+#' @slot t0 (`numeric`)\cr time of initial dosing for each patient.
+#' @slot Tmax (`numeric`)\cr the DLT observation period.
```
danielinteractive, line 307: actually I just checked and `Tmax` should be a scalar. So you can put here `number` in parentheses instead and adjust the validation function below too please

danielinteractive (commented):
thx Wojciech, pls see few comments inside

R/Data-class.R
wwojciech: I was wondering about the same and was about to ask you. Thank you @danielinteractive

R/Data-validity.R
wwojciech: Sure, I was just wondering why in examples they use negative values for `t0`.

R/Data-class.R
```diff
@@ -290,95 +290,61 @@ DataMixture <- function(xshare = numeric(),
...
+#' @param t0 (`numeric`)\cr time of initial dosing for each patient.
+#'   Default to vector of 0s of length equal to length of `u`.
+#' @param Tmax (`number`)\cr the DLT observation period.
+#' @param ... parameters passed to [Data()].
+#'
+#' @export
+#' @example examples/Data-class-DataDA.R
+DataDA <- function(u = numeric(),
+                   t0 = numeric(length(u)),
+                   Tmax = 0 + .Machine$double.xmin,
```
wwojciech, line 341: Hi @danielinteractive maybe I should just set `Tmax = 0` instead of `Tmax = 0 + .Machine$double.xmin` (and the same in corresponding unit tests and default constructor)? What do you think @danielinteractive? I added it because I though it is required that `Tmax` is strictly greater than `0`. But if this is the case, I would probably need to do this for `u` as well.

R/Data-validity.R
wwojciech: @wwojciech Unfortunately, `rolling-crm.Rmd` fails as it has negative `t0` values. What would be you preference then, we allow for negative `t0` or correct `rolling-crm.Rmd`? Actually, I am not so sure, why these negative values were used in that `Rmd` file.....

R/Data-class.R
danielinteractive: @wwojciech hm good question. yeah I think it should be positive.

R/Data-validity.R
```diff
@@ -165,26 +165,17 @@ validate_data_mixture <- function(object) {
 #' contains valid elements with respect to their types, dependency and length.
 validate_data_DA <- function(object) {
   o <- Validate()
+  # In if clause so that below test_* won't fail.
+  if (!test_number(object@Tmax, lower = 0 + .Machine$double.xmin)) {
```
danielinteractive, line 169: unfortunately `test_number` does not allow us to say `strictly greater`. therefore I would here do 
```suggestion
  if (!(test_number(object@Tmax) && object@Tmax > 0)) {
```

danielinteractive: this is a bit cleaner

danielinteractive (approved):

PR #37, opened 2021-12-01
Closes #36

### LINKED ISSUE #36: `as.list` method clean up for `GeneralData`

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #36

### DIFF
--- a/R/Data-methods.R
+++ b/R/Data-methods.R
@@ -1,54 +1,25 @@
-#####################################################################################
-## Author: Daniel Sabanes Bove [sabanesd *a*t* roche *.* com],
-##         Wai Yin Yeung [ w *.*yeung1 *a*t* lancaster *.* ac *.* uk]
-## Project: Object-oriented implementation of CRM designs
-##
-## Time-stamp: <[Data-methods.R] by DSB Mon 11/05/2015 17:43>
-##
-## Description:
-## Methods for handling the data. Plot ideas taken from bcrm package.
-##
-## History:
-## 30/01/2014   file creation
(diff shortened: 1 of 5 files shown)


R/Data-methods.R
```diff
@@ -1,54 +1,26 @@
...
+#' @return a list with all slots in object `x`.
+#'
+#' @example examples/Data-method-asList.R
+#' @export
+#'
+setMethod(
+  f = "as.list",
+  signature = signature(x = "GeneralData"),
+  definition = function(x, ...) {
+    sapply(X = slotNames(x), FUN = slot, object = x, simplify = FALSE)
```
danielinteractive, line 21: `sapply` is discouraged in general, can you use `lapply` and then set names afterwards?

tests/testthat/test-Data-methods.R
```diff
@@ -0,0 +1,9 @@
+# as.list ----
+
+test_that("Coercion creates the expected `List` object", {
```
danielinteractive, line 3:
```suggestion
test_that("Coercion creates the expected list", {
```
(we typically don't use backticks in the test labels)

danielinteractive (approved):
thx Wojciech! only minor things inside, afterwards good to merge

PR #38, opened 2021-12-02
Closes #33

### LINKED ISSUE #33: plot method for `Data` clean up
To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
    - [x] for plot functions/methods use `vdiffr` to compare current test result with original plot
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #33


wwojciech:
This is just a draft which includes plotting function for `Data` with two helpers: `h_data_blind` and `h_tapply`

R/Data-methods.R
```diff
@@ -25,91 +25,93 @@ setMethod(
...
+      # Placebo will be plotted at y = 0 level.
+      x@x[x@x == x@doseGrid[1]] <- 0
+    }
+
+    df <- data.frame(
+      patient = seq_along(x@x),
+      ID = paste(" ", x@ID),
+      dose = x@x,
+      toxicity = as.factor(x@y),
+    )
```
danielinteractive, line 69: please create another helper `h_plot_data_df()` for this, such that you can just call here `df <- h_plot_data_df(x, blind)`. 

For the blinding, I would not do that in `x` directly, the reason is that we don't have a way to signal that this object has been blinded. So I would prefer to keep that at a lower level instead and just do this in the `df`.

R/utils.R
```diff
@@ -0,0 +1,68 @@
...
+#'
+#' @export
+h_data_blind <- function(x) {
+  assert_class(x, "Data")
+  
+  if (x@placebo) {
+    x@x <- h_tapply(x@x, x@cohort, max)
+    x@y <- h_tapply(x@y, x@cohort, sort, decreasing = TRUE)
+  }
+  x
```
danielinteractive, line 67: please only do this in the `df` as mentioned above, probably can just do this then inside the `h_prep_data_df()` thing without separate helper.

R/utils.R
```diff
@@ -0,0 +1,68 @@
...
+#'   This parameter is passed directly into [base::split()] function.
+#' @param FUN (`function`)\cr the function to be applied to each subset of `df` defined by `f`.
+#' @param ... parameters passed to [lapply()], which is used when applying a function `FUN`
+#'   over groups defined by `f`.
+#'
+#' @return The [`data frame`] object with results from `FUN`.
+#'
+#' @export
+#'
+# dapply <- function(df, f, FUN, ...) {
```
danielinteractive, line 23: I would go with this still, as we work on a `df`.

danielinteractive (commented):
thx Wojciech, pls see inside

R/utils.R
```diff
@@ -19,50 +19,14 @@
...
-  assert_class(x, "Data")
-  
-  if (x@placebo) {
-    x@x <- h_tapply(x@x, x@cohort, max)
-    x@y <- h_tapply(x@y, x@cohort, sort, decreasing = TRUE)
-  }
-  x
+  list_df <- split(df, f = f)
+  list_df <- lapply(list_df, FUN, ...)
+  df2 <- do.call(rbind, list_df)
```
danielinteractive, line 29: I think it is a bit better to define the arg list for rbind together with the make row names = FALSE to not get any rownames in the first place

wwojciech: Yap, this is how it was, but then, depending on the output class from `FUN`, the `make row names = FALSE` is treated as a new record to `rbind`, rather than the parameter (for `rbind.data.frame`). See the first example for this function I added to the help page.

wwojciech:
Hmm, do we have any problems with rjags? I see the automation tests fails at the stage of checking rjags dependency.

dinakar29:
> Hi @dinakar29 Do you know if we have any problems with `rjags` package? I see the automation tests fail at the stage of checking `rjags` dependency. Tx!

Yes, there is a minor issue. Let me fix that.

dinakar29:
We should be able to merge https://github.com/Roche/crmPack/pull/39 into `main`, and then merge `main` into this branch to fix the issue.

R/utils.R
danielinteractive: @wwojciech ok thx. then leave that as is.

danielinteractive:
@wwojciech I think you can merge main now into this branch, as @dinakar29 merged his fixes.

tests/testthat/test-helpers.R
```diff
@@ -25,3 +25,33 @@ test_that("h_all_equivalent returns FALSE for non-equivalent objects", {
...
+test_that("h_plot_data_df valid object for sample Data object with placebo and no blinding", {
+  data <- h_get_data()
+  result <- h_plot_data_df(data)
+  expected <- data.frame(
+    patient = 1:12,
+    ID = paste(" ", data@ID),
+    cohort = data@cohort,
+    dose = ifelse(data@x == data@doseGrid[1], 0, data@x),
+    toxicity = as.factor(data@y)
+  )
```
danielinteractive, line 40: it is better to give the explicit values instead of getting them out of `data` - so we can actually see the contents

tests/testthat/test-helpers.R
```diff
@@ -25,3 +25,33 @@ test_that("h_all_equivalent returns FALSE for non-equivalent objects", {
...
+
+  expect_identical(result, expected)
+})
+
+test_that("h_plot_data_df returns valid object for sample Data object with placebo and blinding", {
+  data <- h_get_data()
+  result <- h_plot_data_df(data, blind = TRUE)
+  expected <- data.frame(
```
danielinteractive, line 48: same here

danielinteractive (approved):
thx Wojciech, nice work - just one comment to improve readability of tests, otherwise then good to go.

PR #39, opened 2021-12-02
Update GHA workflows

dinakar29:
Update workflows to install sysdeps, since `jags` is not installed on the image by default and is required by `rjags`.
Also add the validation workflow.


danielinteractive (approved):
thx Dinakar!

PR #40, opened 2021-12-06
Closes #34

### LINKED ISSUE #34: plot method for `DataDual` clean up
To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
    - [x] for plot functions/methods use `vdiffr` to compare current test result with original plot
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #34


R/Data-methods.R
```diff
@@ -33,12 +33,13 @@ setMethod(
...
 #' A method that creates a plot for [`Data`] object.
 #'
-#' @param x (`GeneralData`)\cr object we want to plot.
+#' @param x (`Data`)\cr object we want to plot.
 #' @param y (`missing`)\cr missing object, for compatibility with the generic function.
 #' @param blind (`flag`)\cr indicates whether to blind the data.
 #'   If `TRUE`, then placebo subjects are reported at the same level
 #'   as the active dose level in the corresponding cohort,
 #'   and DLTs are always assigned to the first subjects in a cohort.
+#' @param no_legend (`flag`)\cr should not the legend be added?
```
danielinteractive, line 42:
```suggestion
#' @param legend (`flag`)\cr whether the legend should be added.
```
easier to understand with positive wording of this.

R/Data-methods.R
```diff
@@ -50,7 +51,10 @@ setMethod(
 setMethod(
   f = "plot",
   signature = signature(x = "Data", y = "missing"),
-  def = function(x, y, blind = FALSE, ...) {
+  def = function(x, y, blind = FALSE, no_legend = FALSE, ...) {
```
danielinteractive, line 54:
```suggestion
  def = function(x, y, blind = FALSE, legend = TRUE, ...) {
```

danielinteractive: etc below

tests/testthat/test-Data-methods.R
```diff
@@ -23,3 +23,19 @@ test_that("Plot works as expected for Data object with placebo and with blinding
 
   vdiffr::expect_doppelganger("Plot of Data with placebo and with blinding", result)
 })
+
+# plot-DataDual ----
+
```
danielinteractive, line 28: maybe would be good to add one more test for the `Data` plot method now that you can have the legend switched off?

danielinteractive (commented):
thx Wojciech - nice work, few comments only inside.

PR #47, opened 2021-12-07
Closes #42

### LINKED ISSUE #42: Further streamlining the plot methods for data-like clases.
The cleaning of the plot methods for data-like classes has been completed. Now, when I look at it overall, I think that there are some blocks of plotting code which are used by all three methods and therefore it might be beneficial to consider to extract it into a separate funs. For example: h_plot_data_add_cohort_lines() which would create green lines separating cohorts.

wwojciech:
Closes #42


wwojciech:
Hi @danielinteractive 
I submitted two different commits: `6c64a20` and `d5ef306` with two different solutions. Not sure which one is better. Please let me know what do you think.

danielinteractive:
@wwojciech maybe the second commit is a bit better here. I like this small helper function.

wwojciech:
Thank you @danielinteractive That was my guess in fact :)

wwojciech:
@danielinteractive I will created a help entry and unit tests for this small helper function after you confirm that you accept the interface and etc.

danielinteractive (changes requested):
interface looks good!

danielinteractive (commented):
so good to start with tests etc.

R/helpers.R
```diff
@@ -589,3 +589,41 @@ h_plot_data_df <- function(data, blind = FALSE, ...) {
...
+#'
+#' @param cohort (`integer`)\cr the cohort indices.
+#' @param vertical (`flag`)\cr should the line be vertical? Otherwise it is horizontal.
+#' @param placebo (`flag`)\cr is placebo included in the doses?
+#'   If it so, this function returns `NULL` object as in this case all doses
+#'   in a given cohort are equal and there is no need to separate them.
+#'
+h_plot_data_cohort_lines <- function(cohort,
+                                     vertical = TRUE,
+                                     placebo) {
```
danielinteractive, line 613: better the other way around - arguments that have defaults should be at the end of the function signature

danielinteractive (approved):
thx Wojciech, just one minor thing and then good to merge

PR #48, opened 2021-12-08
Closes #43

### LINKED ISSUE #43: update method for `Data` clean up
To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
  - [x] clean examples
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #43


wwojciech:
Hi @danielinteractive This one is ready for review. I see there are 2 failures in the automation checks at the phase of rendering two vignettes. It is because I added some checking the object validity in the update function. I think this is required, and hence after this update - some problems were revealed with the simulations. I think we may need to discuss this further. Tx!

R/Data-methods.R
```diff
@@ -231,84 +231,75 @@ setMethod(
...
-                  object@cohort <- c(object@cohort,
-                                     rep(max(tail(object@cohort, 1L), 0L),
-                                         length(y)))
-              }
+# Data-update ----
 
-              ## return the object
-              return(object)
-          })
+#' Update method for the [`Data`] class.
```
danielinteractive, line 236:
```suggestion
#' Updating `Data` Objects
```
Title = noun + uppercase style

R/Data-methods.R
```diff
@@ -231,84 +231,75 @@ setMethod(
...
+#' @param object (`Data`)\cr object you want to update.
+#' @param x (`numeric`)\cr the dose level (one level only!).
+#' @param y (`integer`)\cr the DLT vector (0/1 vector) for all patients in this cohort.
+#'   You can also supply `numeric` vectors, but these will then be converted to
+#'   `integer` internally.
+#' @param ID (`integer`)\cr the patient IDs.
+#'   You can also supply `numeric` vectors, but these will then be converted to
+#'   `integer` internally.
+#' @param new_cohort (`flag`)\cr if `TRUE` (default) the new data are assigned to a new cohort.
+#' @param \dots not used.
```
danielinteractive, line 251:
```suggestion
#' @param ... not used.
```
not sure what is better - maybe since we use Rd format here `...` is closer to what the developer sees?

R/Data-methods.R
```diff
@@ -231,84 +231,75 @@ setMethod(
...
+#'
+#' @return The new, updated [`Date`] object.
```
danielinteractive, line 253:
```suggestion
#' @return The new, updated [`Data`] object.
```

R/Data-methods.R
```diff
@@ -231,84 +231,75 @@ setMethod(
...
+  f = "update",
+  signature = signature(object = "Data"),
+  definition = function(object,
+                        x,
+                        y,
+                        ID = length(object@ID) + seq_along(y),
+                        new_cohort = TRUE,
+                        ...) {
+    y_len <- length(y)
+
```
danielinteractive, line 269: would move after the first assertion on `y`

R/Data-methods.R
```diff
@@ -231,84 +231,75 @@ setMethod(
...
+    object@ID <- c(object@ID, safeInteger(ID))
+
+    # Add cohort number.
+    last_cohort <- max(tail(object@cohort, 1L), 0L)
+    object@cohort <- c(
+      object@cohort,
+      rep(last_cohort, y_len) + ifelse(new_cohort, 1L, 0L)
+    )
 
+    stopifnot(test_class(object, "Data"), validObject(object))
```
danielinteractive, line 299:
```suggestion
    validObject(object)
```
should be sufficient as we know the class and this will raise an error if not

danielinteractive (approved):
looks good, thx Wojciech - would be great if we can adjust the example vignette to work afterwards should be good to merge.

PR #49, opened 2021-12-09
Closes #44

### LINKED ISSUE #44: clean update method for `DataParts` class
To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
  - [x] clean examples
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #44


R/Data-methods.R
```diff
@@ -316,60 +316,63 @@ setMethod(
...
+#' @aliases update-DataParts-method
+#' @example examples/Data-method-update-DataParts.R
+#' @export
+#'
+setMethod(
+  f = "update",
+  signature = signature(object = "DataParts"),
+  definition = function(object, check = TRUE, ...) {
+    x <- list(...)$x
+    y <- list(...)$y
```
danielinteractive, line 344: I think it is better to have these as explicit arguments (as it was before)

danielinteractive (commented):
thx Wojciech, one comment otherwise good

wwojciech:
Hi @dinakar29 I am getting the automation check error at SuperLinter phase. Unfortunately, it does not say too much. Could you pl. have a look at this and help me what is wrong here? 
Also, I remember that SuperLinter check is configured here so that it does not report the exact root cause of the problem. Can we somehow adjust it so that the errors reported are more informative for the user? Tx!

R/Data-methods.R
wwojciech: Sure, but then I would need to duplicate the help page for this. As I think inheritsParam for a single parameter wont work here....

wwojciech: Just these (`x`, `y`) or also `ID` as it used to be? If you want me to add the `ID` too, then I would need to initiate it do default as in `update-Data-method` which means duplicated code. And anyway, we pass other arguments to `update-Data-method` via `...`, for example `new_cohort`

danielinteractive: ok can you try please? `@inheritParams` could maybe still work, otherwise can copy those argument descriptions. not a big deal. generally the 3 dots thing we need to be careful with

danielinteractive: the point is arguments we also use in this method should be passed explicitly. other things that only goes upstream can be passed in ...

danielinteractive:
> Hi @dinakar29 I am getting the automation check error at SuperLinter phase. Unfortunately, it does not say too much. Could you pl. have a look at this and help me what is wrong here? Also, I remember that SuperLinter check is configured here so that it does not report the exact root cause of the problem. Can we somehow adjust it so that the errors reported are more informative for the user? Tx!

@wwojciech seems to me that test-Data-methods.R has too long lines. If you look at it in Files Changed tab then you see which ones

R/Data-methods.R
wwojciech: Sure, Tx!

wwojciech:
Thanks @danielinteractive ! I was not aware that SuperLinter error are reported in the Files Changed section :).
The max line size must have been changed in SuperLinter config, since I did not get these errors today morning for the lines which are now highlighted and I have not changed them since then......
Would not be better to set the previous limit for the linter, as it would be quite difficult to short-end all these lines now...?

dinakar29:
> The max line size must have been changed in SuperLinter config

@wwojciech Nothing's changed in the `lintr` configs.

danielinteractive:
@wwojciech please note that the linter only checks the files that were modified. That is, untouched files are not linted. So it can happen that you change 1 character is an old file and suddenly you have 100 linter errors in that file.

wwojciech:
> > The max line size must have been changed in SuperLinter config
> 
> @wwojciech Nothing's changed in the `lintr` configs.

Thanks @dinakar29 How is it possible then that this file passed that check before with these long lines?

danielinteractive (approved):

PR #50, opened 2021-12-09
Closes #45

### LINKED ISSUE #45: clean update method for `DataDual` class
To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
  - [x] clean examples
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #45

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # Version 1.0.9000
-* Updated the `update` method for `Data` class: renamed `newCohort` argument to `new_cohort`,
+* Updated the `update` method for Data-like classes: renamed `newCohort`
+  argument to `new_cohort`,
   introduced validation of the updated object,
   added `check` flag to possibly omit the validation of the updated object.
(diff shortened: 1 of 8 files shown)


wwojciech:
Hi @danielinteractive, do you know how to temporally disable Lintr for the files we have not cleanup yet?
I had to slightly edit `Design-methods.R` to disable validation check for the `update` method and now this whole file wnet under the `Linter` check resulting in hundreds lines of errors....

danielinteractive:
@wwojciech yeah this is the usual R lintr there so you can put `# nolint start` at the start of the file and `# nolint end` at the end. This will exclude the inner region from linting.

danielinteractive (approved):
thx Wojciech

PR #53, opened 2021-12-10
Closes #46

### LINKED ISSUE #46: clean update method for `DataDA` class
To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
  - [x] clean examples
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #46 (I will probably add 2,3 unit tests more for update-DataDA-method, after the PR will be accepted.)


wwojciech:
One issue to be discussed on Monday: `trialtime` and old negative `t0` values. In particular - see the old example for `update-DataDA-method`.

R/Data-class.R
```diff
@@ -82,7 +82,7 @@ NULL
 #'
 #' @note `ID` and `cohort` can be missing, then a warning
 #'   will be issued and the variables will be filled with default
-#'   IDs and best guesses, respectively.
+#'   IDs and best guesses cohort from `\{1, 2, 3, ...\}`.
```
danielinteractive, line 85:
```suggestion
#'   values `1, 2, 3, ...`.
```

R/Data-methods.R
```diff
@@ -276,24 +276,23 @@ setMethod(
                         new_cohort = TRUE,
                         check = TRUE,
                         ...) {
-    assert_number(x)
-    assert_numeric(y, lower = 0, upper = 1, min.len = 1)
+    assert_numeric(x)
+    assert_true(length(x) == 0 || length(x) == 1)
```
danielinteractive, line 280:
```suggestion
    assert_numeric(x, min.len = 0L, max.len = 1L)
```

R/Data-methods.R
```diff
@@ -302,11 +301,15 @@ setMethod(
...
-    last_cohort <- max(tail(object@cohort, 1L), 0L)
-    object@cohort <- c(
-      object@cohort,
-      rep(last_cohort, y_len) + ifelse(new_cohort, 1L, 0L)
-    )
+    if (object@nObs == 0) {
+      new_cohort_ID <- 1L
+    } else {
+      new_cohort_ID <- tail(object@cohort, 1L) + ifelse(new_cohort, 1L, 0L)
+    }
```
danielinteractive, line 308:
```suggestion
    new_cohort_id <- if (object@nObs == 0) {
      1L
    } else {
      tail(object@cohort, 1L) + ifelse(new_cohort, 1L, 0L)
    }
```
(generally better in this kind of patterns) plus avoid upper case in new internal variable names (we keep them when they are already there for external interface to avoid code breaking)

R/Data-methods.R
```diff
@@ -415,6 +417,84 @@ setMethod(
...
+#' @note This function is capable of not only adding new patients but also
+#'   updates existing ones with respect to `y`, `t0`, `u` slots.
+#'
+#' @param object (`DataDA`)\cr object you want to update.
+#' @param u (`numeric`)\cr the new DLT free survival times for all patients,
+#'   i.e. for existing patients in the `object` as well as for new patients.
+#' @param t0 (`numeric`)\cr the time that each patient starts DLT observation window.
+#'   This parameter covers all patients, i.e. existing patients in the `object`
+#'   as well as for new patients.
+#' @param trialtime (`numeric`)\cr current time in the trial, i.e. a followup time.
```
danielinteractive, line 437:
```suggestion
#' @param trialtime (`number`)\cr current time in the trial, i.e. a followup time.
```
as this is only scalar right?

R/Data-methods.R
```diff
@@ -415,6 +417,84 @@ setMethod(
...
+#' @param y (`numeric`)\cr the new DLTs for all patients, i.e. for existing
+#'   patients in the `object` as well as for new patients.
+#' @param ... further arguments passed to `Data` update method [`update-Data-method`].
+#'   These are is used when there are new patients to be added to the cohort.
```
danielinteractive, line 441:
```suggestion
#'   These are used when there are new patients to be added to the cohort.
```

R/Data-methods.R
```diff
@@ -415,6 +417,84 @@ setMethod(
...
+                        t0,
+                        trialtime,
+                        y,
+                        ...,
+                        check = TRUE) {
+    assert_numeric(y, lower = 0, upper = 1)
+    assert_true(length(y) == 0 || length(y) >= object@nObs)
+    assert_numeric(u, lower = 0, len = length(y))
+    assert_numeric(t0, lower = 0, len = length(y))
+    assert_numeric(trialtime, lower = max(object@t0))
```
danielinteractive, line 466:
```suggestion
    assert_number(trialtime, lower = max(object@t0))
```
?

R/Data-methods.R
```diff
@@ -415,6 +417,84 @@ setMethod(
...
+    assert_true(length(trialtime) == 0L || length(trialtime) == 1L)
```
danielinteractive, line 467: I would guess here we can't have an empty `trialtime`?

R/Data-methods.R
```diff
@@ -415,6 +417,84 @@ setMethod(
...
+    # Update slots corresponding to `Data` class.
+    object <- callNextMethod(
+      object = object,
+      y = y[object@nObs + seq_len(n)], # Empty vector when n = 0.
+      ...,
+      check = FALSE
+    )
+
+    # DLT will be observed once the followup time >= the time to DLT.
+    object@y <- safeInteger(y * (trialtime >= t0 + u))
```
danielinteractive, line 482: please adjust comment - we only observe DLT when both time condition is fulfilled and as well in addition `y = 1`

tests/testthat/test-Data-methods.R
```diff
@@ -192,5 +208,31 @@ test_that("Update of DataDual works as expected", {
   object@cohort <- c(object@cohort, 4L, 4L)
 
   expect_valid(result, "DataDual")
-  expect_equal(result, object)
+  expect_identical(result, object)
+})
+
+# update-DataDA ----
+
```
danielinteractive, line 215: if we really need things like missing `trialtime` please make sure to test these edge cases enough, feels pretty risky in terms of code to me

tests/testthat/test-Data-validity.R
```diff
@@ -28,7 +28,7 @@ test_that("validate_subjects returns expected messages for non-valid object", {
     validate_subjects(object),
     c(
       "ID must be of type integer and length nObs and unique",
-      "cohort must be of type integer and length nObs and contain non-negative, sorted values"
+      "cohort must be of type integer and length nObs and contain non-negative, sorted values" # nolintr
```
danielinteractive, line 31: or use paste if too long

tests/testthat/test-Data-validity.R
```diff
@@ -165,7 +165,7 @@ test_that("validate_data_mixture passes for valid object", {
   expect_true(validate_data_mixture(object))
 })
 
-test_that("validate_data_mixture returns error for vector xshare of wrong length", {
+test_that("validate_data_mixture returns error for vector xshare of wrong length", { # nolintr
```
danielinteractive, line 168: maybe check if there is alternative to using `nolintr`

tests/testthat/test-Validate-class.R
```diff
@@ -8,89 +8,85 @@ test_that("Validate objects can be created with default constructor Validate", {
 
 # Validate ----
 
-test_that("Validate returns NULL for valid object, or message for invalid object", {
-  
-  A <- "some_err_A"
-  B <- "some_err_B"
+test_that("Validate returns NULL for valid object, or message for inv. obj.", {
+  erra <- "some_error_A"
+  errb <- "some_error_B"
```
danielinteractive, line 13:
```suggestion
  error_a <- "some_error_A"
  error_b <- "some_error_B"
```
etc below

R/Data-class.R
wwojciech: But the cohort can be like `1,1,1,2,2,3,3,3` which is not "values `1, 2, 3, ...`".

R/Data-methods.R
wwojciech: Tx, how could I missed that! :)

tests/testthat/test-Data-methods.R
wwojciech: I think that it should be `assert_numeric(trialtime, lower = max(object@t0), len = min(length(y), 1))`

wwojciech:
```
assert_numeric(trialtime, lower = max(object@t0), len = min(length(y), 1))
```

R/Data-class.R
```diff
@@ -80,9 +80,9 @@ NULL
 #' @details The `cohort` can be missing if and only if `placebo` is equal to
 #'   `FALSE`.
 #'
-#' @note `ID` and `cohort` can be missing, then a warning
-#'   will be issued and the variables will be filled with default
-#'   IDs and best guesses cohort from `\{1, 2, 3, ...\}`.
+#' @note `ID` and `cohort` can be missing. Then a message will be issued
+#'   and the variables will be filled with default IDs and best guesses cohort,
+#'   i.e. a sorted (in ascending order) sequence of values from `\{1, 2, ...\}`.
```
danielinteractive, line 85: it is a bit too mathematically long wording for my taste but ok ;-)

tests/testthat/test-Data-methods.R
```diff
@@ -236,3 +236,18 @@ test_that("Update of DataDA works as expected", {
   expect_valid(result, "DataDA")
   expect_identical(result, object)
 })
+
```
danielinteractive, line 239: what about the case when there is no real update because empty vectors? can you also add a test for that?

tests/testthat/test-Data-validity.R
```diff
@@ -233,7 +233,7 @@ test_that("validate_data_DA returns error for Tmax of wrong length and negative
   )
 })
 
-test_that("validate_data_DA returns error for vector u of wrong length and values", { # nolintr
+test_that("validate_data_DA: error for u of wrong length and values", {
```
danielinteractive, line 236: ok for here. generally I prefer when there is a meaningful sentence like statement in the tests description instead of a strict schema like "fun: error..." because the former is more readable.

danielinteractive (approved):
thx Wojciech

PR #58, opened 2021-12-14
Closes #54

### LINKED ISSUE #54: `getEff()` method clean up
To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #54


NEWS.md
```diff
@@ -1,4 +1,6 @@
 # Version 1.0.9000
+* Renamed elements in the list object returned by `get_efficacy` function.
+* Renamed `getEff` function to `get_efficacy`.
```
danielinteractive, line 3: please keep original name `getEff` - otherwise we would have to change all camel case function names, which we can't as it breaks too much code. ok to rename some list elements here - generally again we need to be very careful for functions that are used more often like the models etc.

R/Data-methods.R
```diff
@@ -496,54 +496,55 @@ setMethod(
...
+#' @rdname get_efficacy
+#' @aliases get_efficacy-DataDual-method
+#' @example examples/Data-method-get_efficacy.R
+#'
+setMethod(
+  f = "get_efficacy",
+  signature = signature(object = "DataDual"),
+  definition = function(object, ...) {
+    x_DLT <- NULL
+    w_DLT <- NULL
```
danielinteractive, line 534: here it would make lots of sense for the internal variables to avoid uppercase. so use `x_dlt`, `w_dlt` etc. please

R/Data-methods.R
```diff
@@ -496,54 +496,55 @@ setMethod(
...
+    if (any(is_DLT)) {
+      x_DLT <- object@x[is_DLT]
+      w_DLT <- object@w[is_DLT]
+    }
+    if (any(!is_DLT)) {
+      x_no_DLT <- object@x[!is_DLT]
+      w_no_DLT <- object@w[!is_DLT]
     }
+
+    list(x_DLT = x_DLT, w_DLT = w_DLT, x_no_DLT = x_no_DLT, w_no_DLT = w_no_DLT)
```
danielinteractive, line 548: I would put here each element on new line

danielinteractive (commented):
thx Wojciech, pls see inside

danielinteractive (approved):

PR #60, opened 2021-12-16
Closes #55

### LINKED ISSUE #55: `GeneralModel` class clean up
To do:

- [ ] Revise documentation
  - [ ] Replace `##'` by `#'`
  - [ ] Add to `pkgdown.yml`
- [ ] Add unit tests
  - [ ] For classes:
  - [ ] For functions: as usual
- [ ] Refactor as needed
  - [ ] Remove `keywords` tags
  - [ ] For classes:

wwojciech:
Closes #55


tests/testthat/test-Model-validity.R
```diff
@@ -0,0 +1,46 @@
...
+      x
+    },
+    modelspecs = function(x) {
+      x
+    },
+    init = function(x) {
+      x
+    },
+    sample = "param1"
+  )
```
danielinteractive, line 19: maybe put this in another test helper function that gives you a `GeneralModel` object back?

danielinteractive (approved):
thx Wojciech, only minor comments

R/Model-class.R
```diff
@@ -1,99 +1,73 @@
...
+  Class = "GeneralModel",
+  slots = c(
+    datamodel = "function",
+    priormodel = "function",
+    modelspecs = "function",
+    init = "function",
+    sample = "character"
+  ),
+  contains = "AllModels",
+  validity = validate_model_general
```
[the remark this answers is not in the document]
wwojciech: Sure, actually, I had it like you propose and then I changed :)

tests/testthat/test-Model-validity.R
wwojciech: Sure, but I think these 2 places would be only the one we need it, and e.g. for .GeneralData we did not create such helper. Let me know if you still prefer do make it :)

PR #70, opened 2021-12-20
Closes #57

### LINKED ISSUE #57: `LogisticLogNormal` class clean up
To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #57


wwojciech:
Hi @danielinteractive I think that the description of `.LogisticLogNormal` could probably be improved a bit. It is quite difficult for me to do it now, as I do not have yet a full visibility on the usage context of this function.

R/Model-class.R
```diff
@@ -127,144 +127,130 @@ NULL
...
+#' @details The covariate is the natural logarithm of the dose \eqn{x} divided
+#'   by the reference dose \eqn{x^{*}}:
+#'   \deqn{logit[p(x)] = \alpha + \beta \cdot \log(x/x^{*})},
+#'   where \eqn{p(x)} is the probability of observing a DLT for a given dose
+#'   \eqn{x}.
+#'   The prior is: \deqn{(\alpha, \log(\beta)) \sim Normal(\mu, \Sigma)}.
+#'   The slots of this class contain the mean vector and the covariance matrix
+#'   of the bivariate normal distribution, as well as the reference dose.
+#'
+#' @note Te parametrization inside the class uses `alpha0` and `alpha1`.
```
danielinteractive, line 148:
```suggestion
#' @note The parametrization inside the class uses `alpha0` and `alpha1`.
```

R/Model-class.R
```diff
@@ -127,144 +127,130 @@ NULL
...
+#'   `alpha0` is identical to the intercept \eqn{\alpha} above and is the
+#'   log-odds for a DLT at the reference dose `x`*. Therefore, the prior mean
```
danielinteractive, line 150:
```suggestion
#'   log-odds for a DLT at the reference dose `x*`. Therefore, the prior mean
```
and similar below?

R/Model-class.R
```diff
@@ -127,144 +127,130 @@ NULL
...
+#'   for `alpha0` is the expected log-odds at the reference dose `x`* before
+#'   observing any data. Note that the expected odds is not just the exp of the
+#'   prior mean of `alpha0`, because the non-linearity of the exp
+#'   transformation. The log-normal distribution on Wikipedia gives the formula
+#'   for computing the prior mean of `exp(alpha0)`, where `alpha0` is the
+#'   `log(alpha)` in the Neuenschwander et al. (2008) paper. The `alpha1` is
+#'   identical to \eqn{\beta} above and equals to `beta` in the Neuenschwander
+#'   et al paper. The `exp(alpha1)` gives the odds-ratio for DLT between two
+#'   doses that differ by the factor `exp(1)` ~ 2.7. Paramter `alpha1` has a
```
danielinteractive, line 159:
```suggestion
#'   doses that differ by the factor `exp(1)` ~ 2.7. Parameter `alpha1` has a
```

danielinteractive: can you double check the wordlist file that it does not contain wrong spelling terms? thx

R/helpers.R
```diff
@@ -665,3 +665,36 @@ h_check_fun_formals <- function(fun, mandatory = NULL, allowed = NULL) {
 
   mandatory_check && allowed_check
 }
+
+#' Test matrix for positive definiteness
```
danielinteractive, line 669:
```suggestion
#' Testing Matrix for Positive Definiteness
```
(title style)

R/helpers.R
```diff
@@ -665,3 +665,36 @@ h_check_fun_formals <- function(fun, mandatory = NULL, allowed = NULL) {
...
+#'
+#' @description `r lifecycle::badge("experimental")`
+#'
+#' This helper function checks whether a given numeric matrix `x`
+#' is positive-define.
```
danielinteractive, line 674:
```suggestion
#' is positive-definite.
```

danielinteractive (approved):
thx Wojciech, only minor comments inside, good work

R/Model-class.R
wwojciech: Indeed, Tx @danielinteractive , The`WORDLIST` contains wrong `paramter`. I will correct it now!

Issue #73, opened 2021-12-23
Meaning of the `step` argument to be clarified for `McmcOptions` class


wwojciech:
It wold be good to clarify the way we implement `step` feature. That is, if e.g. `step = 2` (without loss of generality, assume that `burnin = 0`, and `iteration = 6`), then a true sample to use would be `1, 3, 5`  or `2, 4, 6`. In other words, do we always start with the first one and then the next, every nth step? or we start from the step-th iteration? 

This question is to make `sample_size`, `save_sample` consistent.

danielinteractive:
I think we should save samples `2,4,6` then. Generally, those samples where (sample - burnin) modulo step is 0.

wwojciech:
Perfect. Thank you @danielinteractive So I will code our `McmcOptions` methods accordingly.

PR #74, opened 2021-12-23
Closes #67

### LINKED ISSUE #67: clean up the 2 methods for `McmcOptions`
See https://github.com/Roche/crmPack/blob/main/R/McmcOptions-methods.R

To do:

- [x] Revise documentation
  - [ ] Replace `##'` by `#'`
  - [ ] Add to `pkgdown.yml`
- [ ] Add unit tests
  - [ ] For classes:
  - [ ] For functions: as usual
- [ ] Refactor as needed
  - [ ] Remove `keywords` tags
  - [ ] For classes:

wwojciech:
Closes #67


wwojciech:
Note: #73

wwojciech:
Hi @dinakar29 would you be having any idea on why this automation check fails at `Docs` phase? Tx!

dinakar29:
> Hi @dinakar29 would you be having any idea on why this automation check fails at `Docs` phase? Tx!

Currently looking into this issue. Will let you know following a resolution.

NAMESPACE
```diff
@@ -106,8 +106,8 @@ export(plotDualResponses)
 export(plotGain)
 export(prob)
 export(probit)
-export(sampleSize)
-export(saveSample)
```
danielinteractive, line 110: please keep the original `camelCase` way of the function names (as mentioned before), as the whole package is like that and I don't want to change too much. thx

R/McmcOptions-methods.R
```diff
@@ -1,51 +1,78 @@
...
+# McmcOptions-sample_size ----
+
+#' @rdname sample_size
+#' @aliases sample_size-McmcOptions-method
+#' @example examples/McmcOptions-methods-sample_size.R
+setMethod(
+  f = "sample_size",
+  signature = signature(object = "McmcOptions"),
+  definition = function(object, ...) {
+    # Iteration numbers relative to object@burnin.
```
danielinteractive, line 35:
```suggestion
```
(clear from code)

R/McmcOptions-methods.R
```diff
@@ -1,51 +1,78 @@
...
+# McmcOptions-save_sample ----
+
+#' @rdname save_sample
+#' @aliases save_sample-McmcOptions-method
+#' @example examples/McmcOptions-methods-save_sample.R
+setMethod(
+  f = "save_sample",
+  signature = signature(object = "McmcOptions"),
+  definition = function(object, iteration, ...) {
+    iteration_relative <- iteration - object@burnin # Relative to object@burnin.
```
danielinteractive, line 75:
```suggestion
    iteration_relative <- iteration - object@burnin
```

danielinteractive (commented):
thx Wojciech, please generally keep function names as before - you can still change / clean up names of internal variables within functions of course. thx

danielinteractive (approved):

dinakar29:
Can't figure out why I cannot reproduce this locally on the same container, but it doesn't work the same on the GH action workflow.

For now, please ignore the pkgdown workflow results until I investigate this further and to unblock progress on other work.

danielinteractive:
Thanks @dinakar29 for looking into it, have a good holiday break first now !

wwojciech:
Thank you Dinakar! @dinakar29 Happy Christmas!

PR #75, opened 2021-12-27
Closes #63

### LINKED ISSUE #63: `dose` method clean up
See https://github.com/Roche/crmPack/blob/b4af53b6585e8a83fc1765cfc9e074732dbd3bfe/R/Model-methods.R#L33

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #63


R/Model-methods.R
```diff
@@ -1,124 +1,124 @@
...
+#'
+setMethod(
+  f = "dose",
+  signature = signature(
+    prob = "numeric",
+    model = "Model",
+    samples = "Samples"
+  ),
+  definition = function(prob, model, samples, ...) {
+    # Extract the `dose` function from the model.
```
danielinteractive, line 48: I would remove these comments as they don't add information

R/Model-methods.R
```diff
@@ -1,124 +1,124 @@
...
+#'
+#' @description `r lifecycle::badge("stable")`
+#'
+#' @param prob (`number`)\cr the toxicity probability which is targeted.
+#' @param model (`Model` or `ModelTox`)\cr the model for single agent dose
+#'   escalation or pseudo DLE (dose-limiting events)/toxicity model.
+#' @param samples (`Samples`)\cr the samples that will be used to compute the
+#'   resulting dose by `model@dose` function.
+#' @param ... not used.
+#' @return A `numeric` output from the `model@dose` function.
```
danielinteractive, line 16: I would explain a bit here that the returned values have one number for each sample.

tests/testthat/test-Model-methods.R
```diff
@@ -0,0 +1,56 @@
...
+  )
+  result <- dose(prob = 0.45, model = dlt_model, samples = samples)
+  expected <- c(0.4075, 0.4515, 0, 0.2998, 0.3321)
+
+  expect_equal(result, expected, tolerance = 0.0001)
+})
+
+# ModelTox_noSamples-dose ----
+
+test_that("ModelTox_noSamples-dose works as expected", {
```
danielinteractive, line 46: Maybe just put the no Samples info into the sentence to make it easier to read

danielinteractive (approved):
Thanks Wojciech, only Minor comments inside

PR #77, opened 2021-12-28
Closes #76

### LINKED ISSUE #76: Question on roxygen aliases for methods
Hi @danielinteractive!

I am wondering if we could consider not adding `-method` suffix for aliases of methods. For example, see `Model-methods.R` file and `@aliases dose-ModelTox-method` alias. I think this suffix does not anything more informative.

I think we use this for consistency with default aliases: e.g: ```\alias{dose,numeric,ModelTox,Samples-method}```

And actually, do you think we could get rid of these automatic aliases in favour of the one we specify manually with `@aliases`

What do you think Daniel?
Thanks a lot!

W.

wwojciech:
Closes #76

### DIFF
--- a/R/Data-methods.R
+++ b/R/Data-methods.R
@@ -1,4 +1,4 @@
-# GeneralData-as.list ----
+# as.list-GeneralData ----
 
 #' Coerce to List Method for `GeneralData` Objects
 #'
@@ -25,7 +25,7 @@ setMethod(
   }
 )
   
-# Data-plot ----
+# plot-Data ----
 
 #' Plot Method for the [`Data`] Class
 #'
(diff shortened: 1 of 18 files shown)


wwojciech:
Hi @dinakar29 , I do apologise for bothering, but since a couple of PR before, The `Roxygen` check fails constantly.
I am cannot reproduce this error locally, and also the error message is not informative enough for me. Could you pl. have a look at this in your free time? Thank you very much!

danielinteractive (approved):
Thanks Wojciech! Re: roxygen checks, as long as you double check that you generated docs ok to merge anyway.

wwojciech:
Thank you @danielinteractive!

PR #84, opened 2022-01-03
Closes #69

### LINKED ISSUE #69: clean up `writeModel`
See https://github.com/Roche/crmPack/blob/dfaf00bc819cb489f5b5f1f519ff07e36cb8f7f4/R/writeModel.R

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [ ] Add unit tests
  - [x] For classes:
  - [ ] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #69


wwojciech:
Hi @danielinteractive I have just pushed my draft PR with a function layout, which I think makes sense. Please let me know fi you accept it so that I can proceed with help change and unit tests.
Thanks a lot!

R/writeModel.R
```diff
@@ -25,62 +23,71 @@
...
+#'  in the BUGS ##' model language, for minor differences see Section Details.
 ##' @param con passed to \code{\link{writeLines}} which actually writes the
 ##' model file
 ##' @param digits number of significant digits used for \pkg{WinBUGS}
 ##' input, see \code{\link{formatC}}
-##' @return Nothing, but as a side effect, the model file is written
 ##'
 ##' @export
-##' @author original idea by Jouni Kerman, modified by Uwe Ligges, DSB removed S
-##' Plus part
```
danielinteractive, line 41: I would remove all this

R/writeModel.R
```diff
@@ -25,62 +23,71 @@
...
-        ## In S-PLUS the source code of a function can be obtained with
-        ## as.character(function_name).
-        ## This omits the "function_name <- function()" piece
-        model.text <- paste("model", as.character(model))
-    }
-    model.text <- gsub("%_%", "", model.text)
-    writeLines(model.text, con = con)
+##' @author original idea by Jouni Kerman, modified by Uwe Ligges,
+##'  DSB removed S-Plus part, WW modified the implementation.
+writeModel <- function(model, con = "model.bug", digits = 5) { # nolintr
```
danielinteractive, line 40:
```suggestion
writeModel <- function(model, file = "model.bug", digits = 5) { # nolintr
```
maybe better argument name?

R/writeModel.R
```diff
@@ -25,62 +23,71 @@
...
+  if (is.R()) {
+    model_text <- c("model", h_replace_scientific(model, digits = digits))
+  } else {
+    # In S-PLUS the body of a function can be obtained with
+    # as.character(function_name).
+    model_text <- paste("model", as.character(model))
+  }
```
danielinteractive, line 47: we always have R, so can be simplified to 
```suggestion
    model_text <- c("model", h_replace_scientific(model, digits = digits))
```

R/writeModel.R
```diff
@@ -25,62 +23,71 @@
...
+    h_format_number,
+    classes = c("integer", "numeric"),
+    digits
+  )
+  bmodel_text <- deparse(bmodel, control = NULL)
+  gsub("\"TEMP_NUM_PREF_|_TEMP_NUM_SUF\"", "", bmodel_text)
+}
+
+# Body apply
+h_bapply <- function(b, f, classes, ...) {
```
danielinteractive, line 67: try to use longer and speaking argument names, e.g. `body`, `fun`

danielinteractive: also assert on the `body` argument - what should it be?

R/writeModel.R
```diff
@@ -25,62 +23,71 @@
...
+  assert_function(f)
+  assert_character(classes, min.len = 1L)
+
+  for (i in seq_along(b)) {
+    if (class(b[[i]]) %in% classes) {
+      b[[i]] <- do.call(f, c(list(b[[i]]), ...))
+    } else if (length(b[[i]]) > 1L) {
+      b[[i]] <- h_bapply(b[[i]], f, classes, ...)
```
danielinteractive, line 75: aha this is recursive

R/writeModel.R
```diff
@@ -25,62 +23,71 @@
...
+  }
+  b
 }
 
+h_format_number <- function(x,
+                            digits = 5,
+                            pref = "TEMP_NUM_PREF_",
+                            suf = "_TEMP_NUM_SUF") {
+  assert_number(x)
+  assert_int(digits)
```
danielinteractive, line 87: also assert `pref` and `suf` (and better call them `prefix` and `suffix`)

R/writeModel.R
```diff
@@ -25,62 +23,71 @@
...
-        bmodel <- gsub(paste('"', i, '"', sep=''), get(i, envir=env), bmodel, fixed=TRUE)
+h_replace_scientific <- function(model, digits = 5) {
+  assert_function(model)
+  assert_int(digits)
+
+  bmodel <- h_bapply(
+    body(model),
+    h_format_number,
+    classes = c("integer", "numeric"),
+    digits
```
danielinteractive, line 60: since you specify the prefix and suffix again in line 63, I would also specify them here, otherwise difficult to keep the code in sync

R/writeModel.R
```diff
@@ -25,62 +23,71 @@
 ##' removed before the BUGS code is saved.
 ##' In S-PLUS, a warning is generated when the model function is defined if the
```
danielinteractive, line 24: remove all S-PLUS mentioning

R/writeModel.R
```diff
@@ -25,62 +23,71 @@
...
+##' be removed before the BUGS code is saved.
 ##'
-##' @param model R function containing the BUGS model in the BUGS
-##' model language, for minor differences see Section Details.
+
+#' @param model R function containing the JAGS model
+#'  in the BUGS ##' model language, for minor differences see Section Details.
 ##' @param con passed to \code{\link{writeLines}} which actually writes the
 ##' model file
 ##' @param digits number of significant digits used for \pkg{WinBUGS}
```
danielinteractive, line 34: I would rather word it for usage with JAGS instead of WinBUGS since we don't support the latter

danielinteractive (commented):
thx Wojciech, direction is good, nice small functions there. see few early comments inside

R/writeModel.R
wwojciech: Thanks Daniel! Actually I had a small 2 problems wit this: 
1. If I use `body` as argument name, I will cover the `body` function from `base` pckg. This of course does not make any harm here, but I am always a bit sceptical about using the same names as well known function names from base packages. For `f` I wanted to be consistent with `rapply` - which was an inspiration for `h_bapply`
2. Actually b can be anything, as `h_bapply` is recursive body apply. So with the first iteration it is just a body of the function but finally can be integer, character or whatsoever.... 
What do you think Daniel?

R/writeModel.R
wwojciech: Yap, I was a bit inspired by `rapply` here.

R/writeModel.R
danielinteractive: thx, re 1: true, but can always use `base::body` if needed for the function. re 2: or just call it `x` if it can be anything?
can also consider calling it `h_rapply` if the difference is small compared to `rapply` and the point is more the recursive application rather than `body`

R/writeModel.R
wwojciech: Sure, so we just ignore the default values of `pref`, `suf` in `h_format_number`?

wwojciech:
Hi @danielinteractive, I have one question: Would you prefer to define our two helper functions for `writeModel` in `writeModel.R` or in `helpers.R`. I think probably `helpers.R`, as after all of these modifications they do not seem to be only `writeModel` specific....

danielinteractive:
@wwojciech yeah agree if not too specific best put in `helpers.R`.
For the helper: I would keep the current name `h_rapply` and still export it. Usually only very few functions should not be exported, vast majority should be exported. This makes it easier for users to extend the package

wwojciech:
Hi Daniel, can you please recommend me the best way of unit testing of functions which write some output into a file? I have some ideas but just wanted to confirm. Tx!

danielinteractive:
@wwojciech good question! I thought someting with `expect_snapshot` can go in the right direction but I am not so sure. The simplest is probably to 
1) create a temporary file
2) write to that file
3) read the file
4) compare with fixed character vector

wwojciech:
Thanks a lot @danielinteractive ! I did exactly as you write and it works!

wwojciech:
@danielinteractive I think we may need to disable Linter for `tests/testthat/_snaps/helpers.md` as this is automatically generated snap file from `expect_snapshot`....

R/helpers.R
```diff
@@ -698,10 +678,191 @@ h_is_positive_definite <- function(x, tolerance = 1e-08) {
...
+  model_text <- c("model", model_text)
+
+  writeLines(model_text, con = file)
+}
+
+#' Recursively Apply a Function to a List
+#'
+#' @description `r lifecycle::badge("experimental")`
+#'
+#' This helper function is conceptually quite similar to [`base::rapply`]
```
danielinteractive, line 780:
```suggestion
#' This helper function is conceptually quite similar to [rapply()]
```
(function is referred to with `()` and `base` does not need to be prefixed)

R/helpers.R
```diff
@@ -698,10 +678,191 @@ h_is_positive_definite <- function(x, tolerance = 1e-08) {
...
+#' function with two major differences (see the note for more details).
+#' It recursively iterates through a "list-like" object and it checks whether
+#' an element is of a given class. If it so, then it replaces that element by
+#' the result of an execution of a given function. Otherwise, and if the element
+#' is of length greater than 1 (i.e. not a scalar), it replaces that element by
+#' the result of `h_rapply`, recursively called for that element. In the
+#' remaining case, that is, the element is not of a given class and is a scalar,
+#' then that element remains unchanged.
+#'
+#' @note This helper function is conceptually similar the same as `rapply`
```
danielinteractive, line 790: duplicate text here, please remove/fix

R/helpers.R
```diff
@@ -698,10 +678,191 @@ h_is_positive_definite <- function(x, tolerance = 1e-08) {
...
+#'   function. However, it differs from `rapply` in two major ways. First, the
+#'   `h_rapply` is not limited to objects of type `list` or `expression` only.
+#'   It can be any "list-like" object of any type for which `[[` operator is
+#'   defined. This can be, for example, an object of type `language`, often
+#'   obtained from the `body` function. The second difference is that the
+#'   flexibility of `rapply` on `how` the result is structured is not available
```
danielinteractive, line 796:
```suggestion
#'   flexibility of `rapply` on how the result is structured is not available
```

danielinteractive: ?

R/helpers.R
```diff
@@ -698,10 +678,191 @@ h_is_positive_definite <- function(x, tolerance = 1e-08) {
...
+#'   with `h_rapply` for the user. That is, with `h_rapply` each element of `x`,
+#'   which has a class included in `classes`, is replaced by the result of
+#'   applying `fun` to the element.
+#'
+#' @param x \cr any "list-like" object for which `[[` operator is defined.
```
danielinteractive, line 801:
```suggestion
#' @param x any "list-like" object for which `[[` operator is defined.
```
(if we don't give class in parenthesis then no need to have line break)

R/helpers.R
```diff
@@ -698,10 +678,191 @@ h_is_positive_definite <- function(x, tolerance = 1e-08) {
...
+#' @param classes (`character`)\cr class names.
+#' @param ... further arguments passed to function `fun`.
+#'
+#' @return "list-like" object of similar structure as `x`.
+#'
+#' @example examples/helpers-rapply.R
+#' @export
+#'
+h_rapply <- function(x, fun, classes, ...) {
+  assert_function(fun)
```
danielinteractive, line 813: maybe assert here that `x[[1]]` works?

R/helpers.R
```diff
@@ -698,10 +678,191 @@ h_is_positive_definite <- function(x, tolerance = 1e-08) {
...
+    }
+  }
+  x
+}
+
+#' Conditional Formatting Using C-style Formats
+#'
+#' @description `r lifecycle::badge("experimental")`
+#'
+#' This helper function conditionally formats a number with [`base::formatC`]
```
danielinteractive, line 830:
```suggestion
#' This helper function conditionally formats a number with [formatC()]
```

R/helpers.R
```diff
@@ -698,10 +678,191 @@ h_is_positive_definite <- function(x, tolerance = 1e-08) {
...
+#' function using `"E"` format and specific number of digits as given by the
+#' user. A number is formatted if and only if its absolute value is less than
+#' `0.001` or greater than `10000`. Otherwise, the number is not formatted.
+#' Additionally, custom prefix or suffix can be appended to character string
+#' with formatted number, so that the changes are marked.
+#' This function is was primarily designed as a helper for [`h_write_model`]
```
danielinteractive, line 836:
```suggestion
#' This function was primarily designed as a helper for [h_write_model()]
```

R/helpers.R
```diff
@@ -698,10 +678,191 @@ h_is_positive_definite <- function(x, tolerance = 1e-08) {
...
+#' function.
+#'
+#' @param x (`number`)\cr a number to be formatted.
+#' @param digits (`function`)\cr the desired number of significant digits.
+#' @param prefix (`string`)\cr a prefix to be added in front of the formatted
+#'   number.
+#' @param suffix (`string`)\cr a suffix to be appended after the formatted
+#'   number.
+#'
+#' @return (`string` or `number`)\cr a formatted `x` or unchanged `x` if the
```
danielinteractive, line 846:
```suggestion
#' @return Either formatted `x` as `string` or unchanged `x` if the
```

R/helpers.R
```diff
@@ -698,10 +678,191 @@ h_is_positive_definite <- function(x, tolerance = 1e-08) {
...
+#'   formatting condition is not met.
+#'
+#' @examples
+#' h_format_number(50000)
+#' h_format_number(50000, prefix = "P", suffix = "S")
+#' @export
```
danielinteractive, line 852: export usually above, not below examples

tests/testthat/test-helpers.R
```diff
@@ -188,16 +188,27 @@ test_that("h_get_slots throws the error for non-existing slots", {
 # h_join_models ----
 
 test_that("h_join_models works as expected", {
-  model1 <- function(x) { x <- x - 2; x <- x^2 } # nolintr
-  model2 <- function(x) { x^3 } # nolintr
+  model1 <- function(x) {
+    x <- x - 2
+    x <- x^2
+  } # nolintr
```
danielinteractive, line 194: I guess we no longer need `nolintr` here - generally we try to have as few as possible of them.

tests/testthat/test-helpers.R
```diff
@@ -214,3 +225,77 @@ test_that("h_join_models throws the error for non-braced expression", {
...
+# h_write_model ----
+
+test_that("h_write_model works as expected", {
+  my_model <- function() {
+    alpha0 <- mean(1:10)
+    alpha1 <- 600000
+  }
+
+  h_write_model(my_model, "./functions_output/h_write_model.jags", 5)
+  expect_snapshot(readLines("./functions_output/h_write_model.jags"))
```
danielinteractive, line 286: please use a temporary file instead (see `?tempfile`), we cannot just write to any directory during tests

danielinteractive (commented):
thx Wojciech, nice work, pls see inside comments - re: lintr let's check with @dinakar29

R/helpers.R
wwojciech: Thanks @danielinteractive I am not so sure what would be the best here. The most reasonable solution which comes to my mind is: `assert_number(invisible(x[[1]]), 1)`. But this is actually obsolete since if x is not subsettable, the error will be thrown from `invisible`, not from the `assert_number` (so the error message will not be that assertion failed). But actually, if `x` is not subsettable, it will anyway fail with the same error at the second line, i.e. ` if (class(x[[i]]) %in% classes)`

danielinteractive: probably we just get a nicer error and user understands better if we check something simple here, e.g. `assert_true(!is.null(x[[1]]))`? but if too difficult, we leave as is, not that critical

wwojciech: yeah, I have tried this approach too, but it will again fail at `x[[1]]` if `x` is not subsettable, without reaching `is.null` fun (see e.g. `assert_true(!is.null(mean[[1]]))` ). Ideally if we have something like `assert_noerror`, but we do not :( So I would leave it as it is.

wwojciech:
The one thing I cannot figure it out is that how to put a reference to `[[` function with a link. Unfortunately, this does not work [`[[`]. So for a moment, I simply left `[[` without a link.

R/helpers.R
wwojciech: The only one idea which comes to my mind at the moment is to use `tryCatch`.

danielinteractive:
> The one thing I cannot figure it out is that how to put with roxygen a reference to `[[` function with a link. Unfortunately, this does not work [`[[`]. So for a moment, I simply left `[[` without a link.

I think you can use something like [`[[`][`Extract`] (the name of the help page is `Extract` but then the link appears as `[[`)

wwojciech:
> thx Wojciech, nice work, pls see inside comments - re: lintr let's check with @dinakar29

Hi @dinakar29 I think that we may need to slightly expand exceptions condition: `          FILTER_REGEX_EXCLUDE: NEWS.md|design/.*` in `linter.yaml`at insightsengineering.

dinakar29:
> > thx Wojciech, nice work, pls see inside comments - re: lintr let's check with @dinakar29
> 
> Hi @dinakar29 I think that we may need to slightly expand exceptions condition: ` FILTER_REGEX_EXCLUDE: NEWS.md|design/.*` in `linter.yaml`at insightsengineering. Maybe just completely ignore whole `_snaps` directory in `testthat`?

Yes, I've added the `_snaps` directory to the exclusion list: https://github.com/insightsengineering/r.pkg.template/blob/main/.github/workflows/linter.yaml#L30

R/helpers.R
danielinteractive: (maybe just have `force(x[[1]])` which evaluates it. if it works, fine, we don't use the result, and otherwise gives clear error here)

danielinteractive (approved):
thx Wojciech!

R/helpers.R
wwojciech: Thanks @danielinteractive What about `invisible(x[[1]])`, as it does not produce any output if no failure, and it produces the error when there is an error?

danielinteractive: also fine, `force` is anyway more a readability thing, it just returns the argument

PR #92, opened 2022-01-11
Closes #91

### LINKED ISSUE #91: add RNG related slots into `McmcOptions` class
Add RNG related slots (kind, seed) into `McmcOptions` class.
The `kind` and `seed` slots should correspond to `.RNG.name` and `.RNG.seed` options available for `inits` parameter in `rjags::jags.model` function. Default values: `kind = "base::Mersenne-Twister"`, `seed = 1`.

wwojciech:
Closes #91

### DIFF
--- a/R/McmcOptions-class.R
+++ b/R/McmcOptions-class.R
@@ -18,14 +18,14 @@ NULL
 #'   For example, for `iterations = 6`, `burnin = 0` and `step = 2`, only
 #'   samples from iterations `2,4,6` will be saved.
 #' @slot rng_kind (`string`)\cr a Random Number Generator (RNG) type used by
-#'   [`rJAGS`]. It must be one out of the following four values:
+#'   [`rjags`]. It must be one out of the following four values:
 #'   `base::Wichmann-Hill`, `base::Marsaglia-Multicarry`,
 #'   `base::Super-Duper`, `base::Mersenne-Twister`, or `NA_character_`.
 #'   If it is `NA_character_` (default), then the RNG kind will be chosen by
-#'   [`rJAGS`].
+#'   [`rjags`].
 #' @slot rng_seed (`number`)\cr a Random Number Generator (RNG) seed
-#'   used by [`rJAGS`] for a chosen `rng_kind`. It must be an integer scalar or
-#'   `NA_integer_`, which means that the seed will be chosen by [`rJAGS`].
+#'   used by [`rjags`] for a chosen `rng_kind`. It must be an integer scalar or
+#'   `NA_integer_`, which means that the seed will be chosen by [`rjags`].
 #'
 #' @aliases McmcOptions
 #' @export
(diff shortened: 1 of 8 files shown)


wwojciech:
Hi @danielinteractive After I submitted this PR, I think that it might be better to have to slots for RNG: type + seed, rather that just one RNG list as it is now....

R/McmcOptions-class.R
```diff
@@ -14,9 +14,20 @@ NULL
 #' @slot step (`count`)\cr only every `step`-th iteration is saved after
 #'   the `burnin`. In other words, a sample from iteration
 #'   `i = 1,...,iterations`, is saved if and only if
-#'   `(i - burnin) mod step = 0`. For example, for `iterations = 6`, 
-#'   `burnin = 0` and `step = 2`, only samples from iterations `2,4,6`
-#'   will be saved.
+#'   `(i - burnin) mod step = 0`.\cr
+#'   For example, for `iterations = 6`, `burnin = 0` and `step = 2`, only
+#'   samples from iterations `2,4,6` will be saved.
+#' @slot RNG (`list`)\cr a Random Number Generator (RNG) type and seed used
```
danielinteractive, line 20: agree with your idea to better make this two slots - easier for the user and we should further abstract here from what internally is needed and passed to JAGS.

R/McmcOptions-class.R
```diff
@@ -14,9 +14,20 @@ NULL
...
+#'   by `rJAGS`. It must be a list with two elements named `.RNG.name` and
+#'   `.RNG.seed`.\cr
+#'   The `.RNG.name` is the the name of the RNG type and it is one out of the
+#'   following four values: `base::Wichmann-Hill`, `base::Marsaglia-Multicarry`,
```
danielinteractive, line 24: I think the user should be able to select from character vector without the `base::` prefix. that is an internal implementation detail we can handle ourselves

R/McmcOptions-class.R
```diff
@@ -44,20 +57,28 @@ NULL
...
 #'
 #' @export
 #' @example examples/McmcOptions-class-McmcOptions.R
 #'
 McmcOptions <- function(burnin = 1e4L,
                         step = 2L,
-                        samples = 1e4L) {
+                        samples = 1e4L,
+                        RNG_kind = "base::Mersenne-Twister",
+                        RNG_seed = 1L) {
```
danielinteractive, line 72: generally here please be careful not to use uppercase in the variable/slot names. so `rng_kind` and `rng_seed` .

danielinteractive (commented):
thx Wojciech, pls see inside - in general, let's try to make it easy for the user - they should not worry about internal implementation details

R/McmcOptions-class.R
wwojciech: Sure, Thanks! I now did it so that it is pasted in the user constructor, so the slots keep the prefix. This is ok, but it has one drawback - the error message from the validity, about wrong prefix....Maybe would be better to keep the names without prefix in slots too, and then have just paste the prefix in `mcmc`...

danielinteractive (approved):

PR #96, opened 2022-01-20
Closes #88

### LINKED ISSUE #88: class `LogisticNormal` clean up
To do:

- [x] Can we share more code with `LogisticLogNormal`?
- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #88 - draft, not sure about dose, prob funs if they should be extracted.


wwojciech:
Hi @danielinteractive I am submitting this PR as a draft, but I think it is quite reasonable proposal. I am only not so sure if `dose`, and `prob` should be extracted as in my proposal. I was also thinking about extracting `modelspecs` and `init` but finally resigned from this idea, as these are quite short. So finally, I apart from `dose` and `prob` I extracted just model functions (i.e. `datamodel` and `priormodel`), which I think makes sense as these are the only `JAGS` model specifications funs.

R/Model-class.R
```diff
@@ -129,6 +131,88 @@ NULL
...
+#'   The prior is: \deqn{(\alpha, \beta) \sim Normal(\mu, \Sigma)}.
+#'   The slots of this class contain the mean vector, the covariance and
+#'   precision matrices of the bivariate normal distribution, as well as the
+#'   reference dose.
+#'
+#' @slot mean (`numeric`)\cr the prior mean vector.
+#' @slot cov (`matrix`)\cr the prior covariance matrix.
+#' @slot prec the prior precision matrix, which is an inverse matrix of the `cov`.
+#' @slot refDose (`number`)\cr the reference dose.
+#'
```
danielinteractive, line 156: maybe make a `@seealso` to the LogisticLogNormal as they are so closely related

R/Model-class.R
```diff
@@ -129,6 +131,88 @@ NULL
...
+    priormodel = jags_model_prior_ln,
+    modelspecs = function() {
+      list(refDose = refDose, priorPrec = prec, priorMean = mean)
+    },
+    init = function() {
+      list(theta = c(0, 1))
+    },
+    sample = c("alpha0", "alpha1"),
+    dose = h_set_env(dose_logistic, environment()),
+    prob = h_set_env(prob_logistic, environment()),
```
danielinteractive, line 208: hm... it would be preferrable if we can do without any environment tricks here.

R/Model-validity.R
```diff
@@ -72,3 +72,16 @@ validate_logistic_log_normal <- function(object) {
...
+
+#' @describeIn validate_model_objects validates that the model parameters
+#'  are valid as well as `refDose` is a positive scalar.
+validate_logistic_normal <- function(object) {
+  result_lln <- result_validate_logistic_log_normal(object)
+
+  o <- Validate()
+  o$check(
+    all.equal(object@cov %*% object@prec, diag(1), check.attributes = FALSE),
+    "prec must be inverse of cov"
```
danielinteractive, line 84: and this here is not required for LogisticLogNormal?

R/Model-validity.R
```diff
@@ -72,3 +72,16 @@ validate_logistic_log_normal <- function(object) {
   )
   o$result()
 }
+
+#' @describeIn validate_model_objects validates that the model parameters
+#'  are valid as well as `refDose` is a positive scalar.
+validate_logistic_normal <- function(object) {
+  result_lln <- result_validate_logistic_log_normal(object)
```
danielinteractive, line 79: hm... I think either
1) have a helper function for this
2) or call the other one directly (`validate_logistic_log_normal`)

R/jags_model.R
```diff
@@ -0,0 +1,39 @@
+# nolint start
+
+# LogisticNormal ----
+
+jags_model_data_ln <- function() {
```
danielinteractive, line 5: in terms of naming, we can here consider a new prefix `jm_`  and then spend more characters on the rest.
e.g. `jm_lik_logistic` (likelihood of logistic model)
note that in the likelihood here it is independent of normal or log-normal prior.

R/jags_model.R
```diff
@@ -0,0 +1,39 @@
...
+  # The logistic likelihood.
+  for (i in 1:nObs) {
+    y[i] ~ dbern(p[i])
+    logit(p[i]) <- alpha0 + alpha1 * StandLogDose[i]
+    StandLogDose[i] <- log(x[i] / refDose)
+  }
+}
+
+jags_model_prior_ln <- function() {
```
danielinteractive, line 14:
```suggestion
jm_prior_logistic_normal <- function() {
```

R/jags_model.R
```diff
@@ -0,0 +1,39 @@
...
+  alpha0 <- theta[1]
+  alpha1 <- theta[2]
+  # Dummy to use `refDose` here. It is contained in the `modelspecs` list
+  # so it must occur also here.
+  # bla <- refDose + 1 # Presumably not needed anymore, WW.
+}
+
+# LogisticLogNormal ----
+
+jags_model_prior_lln <- function() {
```
danielinteractive, line 26:
```suggestion
jm_prior_logistic_log_normal <- function() {
```

R/dose_prob.R
```diff
@@ -0,0 +1,15 @@
+# nolint start
+
+# Logistic ----
+
+dose_logistic <- function(prob, alpha0, alpha1) {
```
danielinteractive, line 5: consider new prefix `d_` and `p_` for these as we will have many different ones

danielinteractive (commented):
thx Wojciech, good start here. if we can do without environment stuff that will be much better.

R/Model-validity.R
wwojciech: Sure, this is a typo here, it should be: `result_lln <- validate_logistic_log_normal(object)`

R/Model-validity.R
wwojciech: Nope, as we do not have `prec` matrix slot for `LogisticLogNormal` class.

R/jags_model.R
wwojciech: I used `data` in the name to somehow indicated this is data model and also to match `datamodel` slot in model related class. I see other logistic models have different formula, so probably would be worth to keep `jm_lik_logistic_normal`, and of course the same will be used to log-normal.

wwojciech:
Hi @danielinteractive I addressed the changes we discussed. However, do be honest I am now not so convinced to extract jags related code into a separate file. This becomes to over-engineered with a log scrutinies, names and rules to remember end etc. Hence, I did `LogisticLogNormal` with the old (but cleaned approach) and `LogisticNormal` with the new one, which as I said do not like too much now :( What do you think?

danielinteractive:
@wwojciech hm ok yeah I agree - I like the `LogisticLogNormal` approach better. so let's do like that.

NAMESPACE
```diff
@@ -278,8 +278,10 @@ importFrom(grid,viewport)
 importFrom(gridExtra,arrangeGrob)
 importFrom(lifecycle,badge)
 importFrom(methods,.valueClassTest)
+importFrom(methods,`slot<-`)
```
danielinteractive, line 281: we can also just import the whole `methods` package. otherwise it gets complicated

R/crmPack-package.R
```diff
@@ -28,14 +28,12 @@
...
+#' @import methods 
 #' @importFrom gridExtra arrangeGrob
 #' @importFrom graphics plot hist legend lines matlines matplot
-#' @importFrom methods setClass setOldClass setGeneric setMethod representation
-#' signature prototype initialize new is .valueClassTest as callNextMethod slot 
-#' slotNames `slot<-` show formalArgs validObject
 #' @importFrom stats binomial coef cov2cor gaussian glm lm median model.matrix
 #' optim pgamma plogis pnorm qgamma qlogis qnorm quantile rbinom rgamma approxfun
-#' rnorm runif uniroot var vcov
+#' rnorm runif uniroot var vcov step
```
danielinteractive, line 36: no need to update this list. during clean up we prefix all stats function calls with `stats::`. therefore at the end we will reduce this list of imported functions to a single function only (only for R CMD check happiness)

tests/testthat/test-Model-class.R
```diff
@@ -68,6 +78,36 @@ test_that("MCMC computes correct values for LogisticNormal model", {
 
 ## dose ----
 
+test_that("dose computes correct values for LogisticNormal model", {
+  data <- h_get_data()
+  model <- h_get_logistic_normal()
+  options <- h_get_mcmc_options(small = TRUE, fixed = TRUE)
+
+  samples <- mcmc(data = data, model = model, options = options)
```
danielinteractive, line 86: I think in this case we don't need real MCMC samples. Instead we are interested if the dose function works correctly and that we can almost read that manually from the test code here.
so I would recommend to construct the samples manually with like 3 different values in it and then create explicit expectation from that and compare.
```suggestion
samples <- Samples(
  data = list(
    alpha0 = c(0, -1, 1),
    alpha1 = c(0, 1, -1)
  ),
  options = McmcOptions(burnin = 0L, step = 1L, samples = 3L)
)
```

tests/testthat/test-Model-class.R
```diff
@@ -68,6 +78,36 @@ test_that("MCMC computes correct values for LogisticNormal model", {
...
+})
+
+## prob ----
+
+test_that("prob computes correct values for LogisticNormal model", {
+  data <- h_get_data()
+  model <- h_get_logistic_normal()
+  options <- h_get_mcmc_options(small = TRUE, fixed = TRUE)
+
+  samples <- mcmc(data = data, model = model, options = options)
```
danielinteractive, line 102: same here

tests/testthat/test-Model-class.R
```diff
@@ -103,3 +142,33 @@ test_that("MCMC computes correct values for LogisticLogNormal model", {
 })
 
 ## dose ----
+
+test_that("dose computes correct values for LogisticLogNormal model", {
+  data <- h_get_data()
+  model <- h_get_logistic_log_normal()
+  options <- h_get_mcmc_options(small = TRUE, fixed = TRUE)
+
+  samples <- mcmc(data = data, model = model, options = options)
```
danielinteractive, line 151: here too

tests/testthat/test-Model-class.R
```diff
@@ -103,3 +142,33 @@ test_that("MCMC computes correct values for LogisticLogNormal model", {
...
+})
+
+## prob ----
+
+test_that("prob computes correct values for LogisticLogNormal model", {
+  data <- h_get_data()
+  model <- h_get_logistic_log_normal()
+  options <- h_get_mcmc_options(small = TRUE, fixed = TRUE)
+
+  samples <- mcmc(data = data, model = model, options = options)
```
danielinteractive, line 167: and here

danielinteractive (approved):
thx Wojciech, just one remaining suggestion, then good to merge.

wwojciech:
Thank you @danielinteractive !

PR #101, opened 2022-01-27
Closes #100

### LINKED ISSUE #100: Add `ModelNormal` parent model class
Add `ModelNormal` parent model class with slots: `mean`, `cov`, `prec` and `ref_dose`.

wwojciech:
Closes #100


NEWS.md
```diff
@@ -1,7 +1,9 @@
 # Version 1.0.9000
+* Renamed `refDose` slot to `ref_dose` in `ModelNormal` class.
+* Created new `ModelNormal` class as a parent class for all the models with
+  bivariate normal prior on the model parameters.
```
danielinteractive, line 4:
```suggestion
  bivariate normal prior on the model parameters and reference dose.
```
?

R/Model-class.R
```diff
@@ -113,90 +121,130 @@ NULL
...
+#'   and the prior \deqn{(alpha0, alpha1) ~ Normal(mean, cov).}
+#'
+#' @seealso [`ModelNormal`] [`LogisticLogNormal`].
+#'
+#' @aliases LogisticNormal
+#' @export
+#'
+.LogisticNormal <- setClass(
+  Class = "LogisticNormal",
+  contains = "ModelNormal",
```
danielinteractive, line 232:
```suggestion
  contains = "ModelNormal"
```
looks nice!

R/Model-class.R
```diff
@@ -207,29 +255,27 @@ LogisticNormal <- function(mean,
...
     prob = function(dose, alpha0, alpha1) {
-      stand_log_dose <- log(dose / refDose)
+      stand_log_dose <- log(dose / ref_dose)
       plogis(alpha0 + alpha1 * stand_log_dose)
     },
-    mean = mean,
-    cov = cov,
-    prec = prec,
-    refDose = refDose
+    params
```
danielinteractive, line 278: I would start with this on top of the construction call.
also maybe rename to `start` or `object` ?

R/Model-class.R
```diff
@@ -324,7 +370,7 @@ LogisticLogNormal <- function(mean,
     datamodel = jags_model_data,
     priormodel = jags_model_prior,
     modelspecs = function() {
-      list(refDose = refDose, cov = cov, mean = mean)
+      list(refDose = refDose, mean = mean, cov = cov)
```
danielinteractive, line 373: hm not sure if this works - because we want to pass the precision matrix here - so that jags does not need to calculate it

R/Model-validity.R
```diff
@@ -58,31 +58,53 @@ validate_logistic_log_normal <- function(object) {
...
+        all.equal(object@cov %*% object@prec, diag(1, 2), check.attributes = FALSE) == TRUE,
+        "prec must be inverse of cov"
+      )
+    }
   }
   o$check(
-    test_number(object@refDose, lower = 0 + .Machine$double.xmin),
-    "refDose must be positive scalar"
+    test_number(object@ref_dose, na.ok = TRUE, lower = 0 + .Machine$double.xmin),
+    "ref_dose must be positive scalar or NA_real_"
```
danielinteractive, line 81: hm I don't think we should allow `NA`. since that will give error in the JAGS then and we want to catch it before.

danielinteractive (commented):
thx Wojciech, good start, few first comments, let me know if you want to chat about any

NEWS.md
wwojciech: but reference dose is not from normal?

R/Model-class.R
wwojciech: I though it would be if the order of arguments corresponds to the order ot classes inheritance.

R/Model-class.R
wwojciech: Yap - that wont work for sure. The `LogisticLogNormal` is not ready yet, I only adjusted the order for one specific reason. I only prepared `LogisticNormal`

R/Model-validity.R
danielinteractive: then I would choose something like `ref_dose = 0`

R/Model-class.R
danielinteractive: order: it is easier to read if it goes on top - then as a reader you see it right away. also inheritance wise is that order - first the parent, then the specific children slots.
name I don't have strong opinion, just for me I think of it as a "start object"

wwojciech: Sure. I have just submitted the new version with the new order. Please let me know if it is ok. Tx!

danielinteractive (approved):
thx Wojciech, looks good, I guess still need few tests for the `ModelNormal` but otherwise good to go.

danielinteractive:
```r
# ModelNormal constructor defines:
modelspecs = function() {
  list(ref_dose = ref_dose, mean = mean, prec = prec)
}
init = function() {
  list(theta = c(0, 1))
}
sample = c("alpha0", "alpha1")
datanames = c("nObs", "y", "x")

.LogisticNormal(
  params,
  datamodel = jags_model_data,
  priormodel = jags_model_prior,
  dose = function(prob, alpha0, alpha1) {
    stand_log_dose <- (logit(prob) - alpha0) / alpha1
    exp(stand_log_dose) * ref_dose
  },
  prob = function(dose, alpha0, alpha1) {
    stand_log_dose <- log(dose / ref_dose)
    plogis(alpha0 + alpha1 * stand_log_dose)
  }
)


# for LogisticLogNormalMixture:

# inherits from LogisticLogNormal.

# inherit:
# mean
# cov
# ref_dose
# prec

# uses LogisticLogNormal() constructor first to get
start <- LogisticLogNormal(...)

# then building on that uses .LogisticLogNormalMixture() constructor overwriting:
modelspecs=
  function(){
    c(
      start@modelspecs(),
      list(catProbs=c(1 - shareWeight, shareWeight))
    )
  }

# overwrite init - because theta is a matrix here and not just vector
init=
  function(){
    list(theta=matrix(c(0, 0, 1, 1), nrow=2))
  }

sample= c(start@sample, "comp")

datanames = c(start@datanames, "nObsshare", "yshare", "xshare")
```

wwojciech:
Thanks @danielinteractive , I have just pushed the changes (for `LogisticNormal` and `LogisticLogNormal` only). Does it look good? Tx!

danielinteractive:
@wwojciech yeah thx looks good, please continue as planned

wwojciech:
Hi @dinakar29 (@danielinteractive  - FYI) Can we disabled `Linter`, `SuperLinter` checks for the whole`vignettes` directory?

PR #102, opened 2022-01-28
85 increments num dose levels extension

### LINKED ISSUE #85: Extension of IncrementsNumDoseLevels method
The existing IncrementsNumDoseLevels method used the last applied dose for determination of allowed increase. We extended the method so that the basis dose level can be either the last applied dose or the maximum given dose.
Two ways of implementation are possible.

1) Extend the current method with an additional parameter. If the parameter is not specified, the default behavior is that the method works as it is now implemented, i.e. increase from the last applied dose. This will ensure backward compatibility with old code.  
2) Implement as a new method.

As we have some other extension, which could either be integrated with approach 1) or 2) it is a general question how future extensions should be integrated.

The attached code contains the implementation with option 1). Code for option 2) can be provided if the decision is to go  this route.
 
[IncrementsNumDoseLevels.R.txt](https://github.com/Roche/crmPack/files/7823137/IncrementsNumDoseLevels.R.txt)

dimitriskontosClinBAY:
Closes #85


danielinteractive:
@dimitriskontosClinBAY I would recommend to 
image
now

R/Rules-class.R
```diff
@@ -367,50 +367,56 @@ IncrementsRelative <- function(intervals,
...
+#' `IncrementsNumDoseLevels`
+#'
+#' @description `r lifecycle::badge("stable")`
+#'
+#' @slot maxLevels (`count`)\cr corresponding to the number of maximum
+#'   dose levels to increment for the next dose. It defaults to 1,
+#' which means that no dose skipping is allowed - the next dose
+#' can be maximum one level higher than the current dose.
+#' @slot basisLevel (`string`)\cr corresponding to the dose level used to increment from.
+#' It can take two possible values 'lastGiven' or 'maxGiven'. If lastGiven (default)
```
danielinteractive, line 383:
```suggestion
#' It can take two possible values `lastGiven` or `maxGiven`. If `lastGiven` (default)
```
also below - the reason is that 1) this formats the string as code, 2) the spell check is not run on this text (because it is recognized as code).

R/Rules-class.R
```diff
@@ -367,50 +367,56 @@ IncrementsRelative <- function(intervals,
...
+#'
+#' @export
+#' @example examples/Rules-class-IncrementsNumDoseLevels.R
+#' 
+IncrementsNumDoseLevels <- function(maxLevels=1,
+                                    basisLevel="lastGiven"){
+  .IncrementsNumDoseLevels(
+    maxLevels=safeInteger(maxLevels),
+    basisLevel=basisLevel
+    )
```
danielinteractive, line 416:
```suggestion
  )
```

R/Rules-methods.R
```diff
@@ -744,35 +730,50 @@ setMethod("maxDose",
...
-
-              ## so the maximum next dose is
-              ret <- data@doseGrid[maxNextDoseLevel]
-
-              return(ret)
-            })
+# maxDose-IncrementsNumDoseLevels ----
 
+#' @rdname maxDose
+#'
```
danielinteractive, line 738: just need to delete the one blank to satisfy the linter check
```suggestion
#'
```
similar below

R/Rules-methods.R
```diff
@@ -744,35 +730,50 @@ setMethod("maxDose",
...
+setMethod(
+  "maxDose",
+  signature = signature(
+    increments = "IncrementsNumDoseLevels",
+    data = "Data"
+  ),
+  definition  = function(increments, data, ...) {
+    # Determine what is the basis level for increment,
+    # i.e. the last dose or the max dose applied.
+    basisDoseLevel <- ifelse(
```
danielinteractive, line 757: use snake case for internal variables, so here
```suggestion
    basis_dose_level <- ifelse(
```

danielinteractive (commented):
thx Dimitris, that looks great! Thanks a lot for the good work. See few comments inside.
Also in addition, can you please restore/not delete the snapshot files? (in the `_snaps` subdirectory) It is not your fault - sometimes these are deleted because the tests don't reach these plot comparisons and then testthat thinks they should be deleted. let me know if you need help with that.

danielinteractive (commented):
thx @dimitriskontosClinBAY , can you fix the remaining linter issues? see the "Files changed" tab

danielinteractive (approved):

PR #105, opened 2022-02-03
103: Make `dose` and `prob` usual S4 methods

### LINKED ISSUE #103: Move `dose` `prob` functions from Class slots to Class methods
Move `dose` `prob` functions from Class' slots to Class' methods for the following model classes which were already cleaned up: `LogisticNormal`, `LogisticLogNormal`, `LogisticLogNormalSub`, `ProbitLogNormal`, `ProbitLogNormalRel`. The slots `dose` and `prob` should not be used anymore.

wwojciech:
Closes #103


R/helpers.R
```diff
@@ -845,3 +845,54 @@ h_validate_combine_results <- function(v1, v2) {
...
+#' @param dose (`number` or `numeric`)\cr the dose to be validated.
+#' @param is_sample_scalar (`flag`)\cr is model parameter sample scalar?
+#'
+#' @export
+#' @examples
+#' h_assert_dose(c(34, 12), TRUE)
+#' \dontrun{
+#' h_assert_dose(c(34, 12), FALSE) # Error.
+#' }
+h_assert_dose <- function(dose, is_sample_scalar) {
```
danielinteractive, line 892: I would avoid two assertion functions here.
rather in the dose and prob methods use a combination of
1) checking range of dose/ prob - that is the difference here
and then
2) check that either dose/prob or samples are scalar. not even sure if we need separate assertion function for this.
something like `assert_true(test_scalar() || test_scalar())` could work too.

danielinteractive (commented):
thx Wojciech, generally good and what we discussed, just please consider major comment to simplify things. we can discuss briefly later and then decide

NEWS.md
```diff
@@ -1,4 +1,6 @@
 # Version 1.0.9000
+* Added new `doseFunction` and `probFunction` functions which return dose
+  (or probability) computing functions using model specific parameters.
```
danielinteractive, line 3:
```suggestion
  and probability computing functions using model specific parameters.
```

danielinteractive (approved):
thx Wojciech!

Issue #152, opened 2022-03-10
Verify whether we can use in JAGS (datamodel) body function code conditionally.


wwojciech:
Example from `DualEndpoint` constructor (but this is also valid for some other classes):
```
  datamodel <- if (use_log_dose) {
    function() {
      for (i in 1:nObs) {
        # The toxicity model.
        meanZ[i] <- betaZ[1] + betaZ[2] * log(x[i] / ref_dose)
        z[i] ~ dnorm(meanZ[i], 1)
        y[i] ~ dinterval(z[i], 0)

        # The conditional biomarker model. betaW defined in subclasses!
        condMeanW[i] <- betaW[xLevel[i]] + rho / sqrt(precW) * (z[i] - meanZ[i])
        w[i] ~ dnorm(condMeanW[i], condPrecW)
      }
    }
  } else {
    function() {
      for (i in 1:nObs) {
        # The toxicity model.
        meanZ[i] <- betaZ[1] + betaZ[2] * (x[i] / ref_dose)
        z[i] ~ dnorm(meanZ[i], 1)
        y[i] ~ dinterval(z[i], 0)

        # The conditional biomarker model. betaW defined in subclasses!
        condMeanW[i] <- betaW[xLevel[i]] + rho / sqrt(precW) * (z[i] - meanZ[i])
        w[i] ~ dnorm(condMeanW[i], condPrecW)
      }
    }
  }
```

Currently, this above code seems quite redundant. I was wondering whether, it could be simplified to something like:

```
datamodel <- function() {
  if (use_log_dose) {
    transform <- log
  } else {
    transform <- I
  }
  for (i in 1:nObs) {
    # The toxicity model.
    meanZ[i] <- betaZ[1] + betaZ[2] * transform(x[i] / ref_dose)
    z[i] ~ dnorm(meanZ[i], 1)
    y[i] ~ dinterval(z[i], 0)
    
    # The conditional biomarker model. betaW defined in subclasses!
    condMeanW[i] <- betaW[xLevel[i]] + rho / sqrt(precW) * (z[i] - meanZ[i])
    w[i] ~ dnorm(condMeanW[i], condPrecW)
  }
}
```

wwojciech:
@danielinteractive - FYI

danielinteractive:
[...]

PR #153, opened 2022-03-10
Closes #144

### LINKED ISSUE #144: model class `DualEndpoint` clean up

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [ ] Add unit tests
  - [ ] For classes:
  - [ ] For functions: as usual
- [ ] Refactor as needed
  - [ ] Remove `keywords` tags
  - [ ] For classes:

wwojciech:
Closes #144


wwojciech:
Hi @danielinteractive This is just a draft - without some unit tests yet. The user construction function is still a bit long - but maybe I may have some idea to make it shorter - see #152

R/Model-class.R
```diff
@@ -931,338 +931,224 @@ LogisticLogNormalMixture <- function(mean,
...
+#'   slot `sigma2W`.
+#'
+#'   Finally, the two endpoints \eqn{y} (the binary DLT variable) and \eqn{w}
+#'   (the biomarker) can be correlated, by assuming a correlation of level
+#'   \eqn{rho} between the underlying continuous latent toxicity variable \eqn{z}
+#'   and the biomarker \eqn{w}. Again, this correlation can be fixed or assigned
+#'   a prior distribution from the scaled Beta family; see the details below
+#'   under slot `rho`.
+#'
+#'   Please see the Hive page for more details on the model and the example
```
danielinteractive, line 975:
```suggestion
#'   Please see the example
```

R/Model-class.R
```diff
@@ -931,338 +931,224 @@ LogisticLogNormalMixture <- function(mean,
...
-
-    priormodel <-
-        function(){
-            ## priors for betaW: defined in subclasses
-
-            ## the bivariate normal prior for the
-            ## probit coefficients
-            log.betaZ[1:2] ~ dmnorm(mu[], PrecBetaZ[,])
+                         rho) {
+  use_fixed <- sapply(list(sigma2W = sigma2W, rho = rho), is.scalar)
```
danielinteractive, line 1054: I would write this more explicit, since it is only 2 elements, e.g. 
```suggestion
  use_fixed <- list(sigma2W = is.scalar(sigma2W), rho = is.scalar(rho))
```

R/Model-class.R
```diff
@@ -931,338 +931,224 @@ LogisticLogNormalMixture <- function(mean,
...
+#'   prior parameters.
+#' @param rho (`numeric`)\cr either a fixed value for the correlation
+#'   (between -1 and 1), or a named vector with elements named `a` and `b` for
+#'   the Beta prior on the transformation `kappa = (rho + 1) / 2`, which is in
+#'   (0, 1). For example, `a = 1, b = 1` leads to a uniform prior on `rho`.
+#'
+#' @export
+#'
+DualEndpoint <- function(mean,
+                         cov,
```
danielinteractive, line 1049: I wonder here since I am not too worried about backwards compatibility whether we should directly take the `ModelParamsNormal` instead of separate `mean` and `cov`?

R/Model-validity.R
```diff
@@ -157,3 +148,61 @@ v_model_logistic_log_normal_mix <- function(object) {
...
+      h_test_named_numeric(object@rho, permutation.of = c("a", "b")),
+      "rho must be a named numerical vector of length two with positive finite values and names 'a', 'b'"
+    )
+  }
+
+  v$result()
+}
+
+#' @describeIn v_model_objects validates that [`DualEndpointRW`] class slots are valid.
+v_model_dual_endpoint_RW <- function(object) {
```
danielinteractive, line 206: please only use lower case in function names generally
```suggestion
v_model_dual_endpoint_rw <- function(object) {
```

danielinteractive (commented):
Thanks @wwojciech , yes, your proposal makes sense - basically these helper functions modify the 4 components, so take them as input (in addition to other stuff) and give them back as output.

R/Model-class.R
wwojciech: Sure, we could do this, but that would not only be a backward compatibility issue (which is not the case as you say) but also inconsistency with other (normal) Model constructors.....

danielinteractive: ok, then let's keep as is

R/helpers_model.R
```diff
@@ -0,0 +1,94 @@
...
+#' A simple helper function that takes [`DualEndpoint`] model existing components
+#' (`priormodel`, `modelspecs`, `init`, `sample`), and updates them with regard to
+#' DLT and biomarker correlation `rho`.
+#'
+#' @param use_fixed (`flag`)\cr TODO
+#' @param rho (`numeric`)\cr TODO
+#' @param priormodel (`function`)\cr TODO
+#' @param modelspecs (`function`)\cr TODO
+#' @param init (`function`)\cr TODO
+#' @param sample (`character`)\cr TODO
```
danielinteractive, line 62: remember then that you can `@inheritParams` from other function's doc here

wwojciech: Sure, thanks for reminder!

danielinteractive (approved):

wwojciech:
Hi @dinakar29 , - can you please help me on why validation check is failing here are RMD Check. I run it locally - no error, also the error visible here is not informative. Tx!

danielinteractive:
@dinakar29 adding detail for this question. 
image

wwojciech:
Seems that this is some problem in generating manual in the pdf format:

```
> path
[1] "/Volumes/Roche/NEST"
> name
[1] "crmPack_1.0.9000.pdf"
> pkg$path
/Volumes/Roche/NEST/crmPack
> callr::rcmd("Rd2pdf", cmdargs = c("--force", paste0("--output=", path, "/", name), pkg$path), fail_on_status = TRUE, stderr = "2>&1", spinner = FALSE)
Error in (function (command = NULL, args = character(), error_on_status = TRUE,  : 
  System command 'R' failed, exit status: 1, stdout + stderr (last 10 lines):
E>  []\T1/zi4/m/n/9 ## Obtain prior modal estimates for the Effloglog model (effic
E> acy model) given the pseudo data. 
E> [51]
E> Overfull \hbox (12.73181pt too wide) in paragraph at lines 2814--2814
E>  []\T1/zi4/m/n/9 ## specifying the shape (a) and the rate (b) parameters for th
E> e gamma distribution in (nu slot).[] 
E> 
E> Overfull \hbox (5.28088pt too wide) in paragraph at lines 2817--2817
E>  []\T1/zi4/m/n/9 Effmodel<-Effloglog(Eff=c(1.223,2.513),Effdose=
E> Error in running tools::texi2pdf()
Type .Last.error.trace to see where the error occurred
> .Last.error.trace

 Stack trace:

 1. callr::rcmd("Rd2pdf", cmdargs = c("--force", paste0("--output=",  ...
 2. callr:::run_r(options)
 3. base:::with(options, with_envvar(env, do.call(processx::run,  ...
 4. base:::with.default(options, with_envvar(env, do.call(processx::run,  ...
 5. base:::eval(substitute(expr), data, enclos = parent.frame())
 6. base:::eval(substitute(expr), data, enclos = parent.frame())
 7. callr:::with_envvar(env, do.call(processx::run, c(list(bin, args = real_cmdargs,  ...
 8. base:::force(code)
 9. base:::do.call(processx::run, c(list(bin, args = real_cmdargs,  ...
 10. (function (command = NULL, args = character(), error_on_status = TRUE,  ...
 11. throw(new_process_error(res, call = sys.call(), echo = echo,  ...

 x System command 'R' failed, exit status: 1, stdout + stderr (last 10 lines):
E>  []\T1/zi4/m/n/9 ## Obtain prior modal estimates for the Effloglog model (effic
E> acy model) given the pseudo data. 
E> [51]
E> Overfull \hbox (12.73181pt too wide) in paragraph at lines 2814--2814
E>  []\T1/zi4/m/n/9 ## specifying the shape (a) and the rate (b) parameters for th
E> e gamma distribution in (nu slot).[] 
E> 
E> Overfull \hbox (5.28088pt too wide) in paragraph at lines 2817--2817
E>  []\T1/zi4/m/n/9 Effmodel<-Effloglog(Eff=c(1.223,2.513),Effdose=
E> Error in running tools::texi2pdf() 

> 
```

wwojciech:
Ok, I found it (https://github.com/Roche/crmPack/issues/159). Such small thing but so difficult to spot it :(

PR #161, opened 2022-03-21
Closes #145

### LINKED ISSUE #145: model class `DualEndpointRW` clean up

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #145


R/Model-class.R
```diff
@@ -1202,12 +1201,14 @@ DualEndpoint <- function(mean,
   )
   # Update model components with regard to biomarker regression variance.
   comp <- h_model_dual_endpoint_sigma2W(
-    use_fixed["sigma2W"], sigma2W = sigma2W, comp = comp
+    use_fixed["sigma2W"],
+    sigma2W = sigma2W, comp = comp
```
danielinteractive, line 1205:
```suggestion
    sigma2W = sigma2W, 
    comp = comp
```

R/Model-class.R
```diff
@@ -1202,12 +1201,14 @@ DualEndpoint <- function(mean,
...
   )
 
   # Update model components with regard to DLT and biomarker correlation.
   comp <- h_model_dual_endpoint_rho(
-    use_fixed["rho"], rho = rho, comp = comp
+    use_fixed["rho"],
+    rho = rho, comp = comp
```
danielinteractive, line 1211:
```suggestion
    rho = rho, 
    comp = comp
```

R/Model-class.R
```diff
@@ -1230,227 +1231,142 @@ DualEndpoint <- function(mean,
...
-
-    } else {
-        start@modelspecs <-
-            function(){
-                c(oldModelspecs(),
-                  list(precBetaW=1/sigma2betaW))
-            }
-    }
+  start@priormodel <- if (rw1) {
+    h_jags_join_models(
```
danielinteractive, line 1321: since priormodel is being joined in both if and else blocks, how about first just assigning the rw part (with if / else) and then doing the join?

R/Model-class.R
```diff
@@ -1230,227 +1231,142 @@ DualEndpoint <- function(mean,
...
-    ## inherit everything from DualEndpoint object "start", and add
-    ## specifics
-    .DualEndpointRW(start,
-                    sigma2betaW=sigma2betaW,
-                    useRW1=useRW1)
+  start@sample <- c(start@sample, "betaW", "delta")
+  # We need the doseGrid in the BUGS model.
+  start@datanames <- c(start@datanames, "doseGrid")
 
+  # Update model components with regard to biomarker regression variance.
```
danielinteractive, line 1355: nice!

danielinteractive (commented):
thx Wojciech, pls see inside few minor comments

R/Model-class.R
wwojciech: Sure, great idea, Tx!

danielinteractive (approved):

PR #170, opened 2022-03-31
Closes #168

### LINKED ISSUE #168: ref_dose proper validation
For some models it is required that `ref_dose >= 0`, and for some other`ref_dose > 0`. This should be reflected in validation functions.

We have at least two options to consider:

```
# option 1
.A <- setClass("A", slots = c(a = "character")) # ModelLogNormal, ....
.B <- setClass("B", slots = c(b = "logical")) # ref dose - RefDosePositive, RefDoseNonNegative
.C <- setClass("C", contains = c("A", "B"), slots = c(c = "factor"))

foo <- .C(a = "bla", b = FALSE, c = factor("bla"))
foo
showClass("C")

# option 2
library(checkmate)
.positive_number <- setClass("positive_number", contains = "numeric", validity = function(object) {test_number(object, lower = 0)} )
# in the user constructor for D:
.positive_number(1)
.positive_number(-1)

# so that the user does not need to do it

.D <- setClass("D", contains = "A", slots = c(d = "factor", ref_dose = "positive_number"))
```

wwojciech:
Closes #168


R/helpers.R
```diff
@@ -42,6 +42,31 @@ Validate <- setRefClass(
...
+#'
+#' @name positive_number
+#'
+positive_number <- setClass(
+  Class = "positive_number",
+  contains = "numeric",
+  validity = function(object) {
+    v <- Validate()
+    v$check(
+      test_number(object, lower = .Machine$double.xmin, finite = TRUE),
```
danielinteractive, line 63: actually how about simplifying this:
```suggestion
      test_number(object, finite = TRUE) && object > 0,
```
seems more explicit and does the same

danielinteractive (approved):
thx Wojciech, just one suggestion inside

R/helpers.R
wwojciech: Sure, we can do that in this case as `a && b`, `b` is not evaluated if `a` `FALSE`, so we are sure that `object > 0` does not throw any error. Would you like to apply this approach also for other places?

danielinteractive: if we need this more than once (here), then I would suggest to do our `test_number_limits()` function so we can use it here e.g. with `lower = 0, include_lower = FALSE` arguments here

PR #198, opened 2022-04-28
Closes #187

### LINKED ISSUE #187: model class `Effloglog` cleanup

To do:

- [ ] Revise documentation
  - [ ] Replace `##'` by `#'`
  - [ ] Add to `pkgdown.yml`
- [ ] Add unit tests
  - [ ] For classes:
  - [ ] For functions: as usual
- [ ] Refactor as needed
  - [ ] Remove `keywords` tags
  - [ ] For classes:

wwojciech:
No unit tests yet.


R/Model-class.R
```diff
@@ -1938,239 +1936,227 @@ LogisticIndepBeta <- function(binDLE,
...
+  nobs_no_dlt <- length(eff_obsrv)
+  if (nobs_no_dlt > 0L) { # Observed data available.
+    # Set X, Y to observed data only.
+    X <- model.matrix(fit_eff)[-seq_along(eff), ]
+    Y <- eff_obsrv
+
+    fit_eff0 <- lm(eff ~ log(log(eff_dose))) # Pseudo only.
+    X0 <- model.matrix(fit_eff0)
+    mu0 <- coef(fit_eff0)
+    Q0 <- t(X0) %*% X0
```
danielinteractive, line 2132: same here

R/mcmc.R
```diff
@@ -396,24 +396,22 @@ setMethod("mcmc",
                      ...){
 
               ## decide whether we sample from the prior or not
-              from_prior <- data@nObs == 0L
+              from_prior <- data@nObs == 0L # TODO shouldn't be no DLT only?
```
danielinteractive, line 399: maybe check with @Winnie-Yeung

danielinteractive (commented):
thx Wojciech! looks good so far

R/mcmc.R
wwojciech: Sure, Tx!

wwojciech: Winnie does not really remember about how this parameter can be used in this function, so I just removed it.

wwojciech:
Hi @dinakar29 Do you know how we can trigger the checks here? I have tried to do this by using the `Re-run all` button in the actions - but it did not start...
Maybe this is because I first created a draft PR and then marked it as ready for review? But I think that still this should not happen and the checks should be started when I marked this draft PR as ready for review. 
Thank you for your help!

W.

danielinteractive (approved):

inst/WORDLIST
```diff
@@ -109,6 +109,7 @@ DualSimulationsSummary
 EffFlexi
 Effloglog
 Effmodel
+efficacies
```
danielinteractive, line 112: is that a real word?... maybe better use `efficacy values`  in text?

danielinteractive (changes requested):

inst/WORDLIST
wwojciech: I checked in the dictionary and `efficacies` is ok but I can change it to `efficacy values`.

danielinteractive (approved):

PR #199, opened 2022-05-05
Closes #188

### LINKED ISSUE #188: model class `EffFlexi` cleanup

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #188


R/Model-class.R
```diff
@@ -2156,189 +2156,174 @@ Effloglog <- function(eff,
...
+EffFlexi <- function(eff,
+                     eff_dose,
+                     sigma2W,
+                     sigma2betaW,
+                     rw1 = TRUE,
+                     data) {
+  assert_numeric(eff)
+  assert_numeric(eff_dose)
+  assert_flag(rw1)
+
```
danielinteractive, line 2291: also need to assert `data`, potentially also `sigma2W` and `sigma2betaW`

danielinteractive (approved):
thx Wojciech! lots of work here.

R/Model-class.R
wwojciech: Tx @danielinteractive, you mean `assert_true(validObject(data))`for `data`?
I am wondering about `sigma2W` and `sigma2betaW` (and maybe also about `data`), as we do no not use it much in the bodu of the user constructor, and these objects are later validated by the class validation function. What do you think?

danielinteractive: I would just say `assert_class(data, "Data")` or what the class should be.
For the sigmas, as soon as we touch it within the constructor, then I would assert it (just `numeric`). otherwise not.

PR #209, opened 2022-05-19
Closes #208

### LINKED ISSUE #208: model class `TITELogisticLogNormal` cleanup

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #208


tests/testthat/test-Model-class.R
```diff
@@ -853,3 +853,75 @@ test_that("MCMC computes correct values for DALogisticLogNormal model and empty
...
+  result <- mcmc(data = data, model = model, options = options)
+  expect_snapshot(result@data)
+})
+
+test_that("MCMC computes correct values for TITELogisticLogNormal model (linear, empty data)", {
+  data <- h_get_data_da(empty = TRUE)
+  model <- h_get_tite_logistic_log_normal("linear")
+  options <- h_get_mcmc_options(small = TRUE, fixed = TRUE)
+
+  # Suppress JAGS warnings about unused variables `zeros` and `cadj`.
```
wwojciech, line 896: Hi @danielinteractive - I will open a separate issue for this, as it requires some global changes. This "unused variables" JAGS warning is because `model@modelspecs` returns all of the parameters, regardless of the whether the (merged `datamodel`, `priormodel`) JAGS model use it or not. And in this case, the `priormodel` dose not use `zeros` and `cadj` vars. Probably we may need to parametrize `model@modelspecs` function with some additional argument, such as `from_prior`.

danielinteractive: thx @wwojciech , sounds good - either reparametrize or similar as with the `datanames` add another slot that could be used if needed only? we can discuss then more in the other issue.

danielinteractive (approved):
very nice, thanks Wojciech!!

PR #217, opened 2022-05-24
Closes #216

### LINKED ISSUE #216: Update of `modelspec` slot of `GeneralModel`.
It would be great if `modelspec` function (slot) in `GeneralModel` is parametrized by some flag parametr so that its output can depend on the value of that flag. We need this because, we distinguish between `datamodel` and `priormodel` and therefore, there are some scenarios in which some parameters are required only for `datamodel` and they are not required by the `priormodel`, hence it should not be supplied with `modelspec`, if `datamodel` is not used (in case of empty data for instance).
This flag should be of a fixed name and should be interpreted by `mcmc` methods (`h_jags_get_data` in case of `mcmc` with `GeneralData`, `GeneralModel` signature).
The problem behind this issue is a bit connected with https://github.com/Roche/crmPack/issues/206

So far, we had to suppress the warnings (about unused variables) thrown by the JAGS. See unit test examples below:

```
  data <- h_get_data_da(empty = TRUE)
  model <- h_get_tite_logistic_log_normal("linear")
  options <- h_get_mcmc_options(small = TRUE, fixed = TRUE)

  # Suppress JAGS warnings about unused variables `zeros` and `cadj`.
  result <- suppressWarnings(mcmc(data = data, model = model, options = options))
  expect_snapshot(result@data)
```
or

```
  data <- h_get_data(empty = TRUE)
  model <- h_get_one_par_exp_normal_prior()
  options <- h_get_mcmc_options(small = TRUE, fixed = TRUE)

  suppressWarnings(result <- mcmc(data = data, model = model, options = options))
  expect_snapshot(result@data)
```

- [x] Parametrize `modelspec` for all the model classes. Update help pages.
- [x] Update `mcmc` methods to handle new parametrized `modelspec`.
- [x] Remove `suppressWarnings` from the unit tests.

wwojciech:
Closes #216

### DIFF
--- a/R/Model-class.R
+++ b/R/Model-class.R
@@ -198,11 +198,11 @@ ModelLogNormal <- function(mean, cov, ref_dose = 1) {
       alpha1 <- exp(theta[2])
     },
     modelspecs = function(from_prior) {
-      if (from_prior) {
-        list(mean = params@mean, prec = params@prec)
-      } else {
-        list(ref_dose = ref_dose, mean = params@mean, prec = params@prec)
+      ms <- list(mean = params@mean, prec = params@prec)
+      if (!from_prior) {
+        ms$ref_dose <- ref_dose
       }
+      ms
     },
     init = function() {
       list(theta = c(0, 1))
@@ -384,11 +384,11 @@ LogisticLogNormalSub <- function(mean, cov, ref_dose = 0) {
       alpha1 <- exp(theta[2])
     },
     modelspecs = function(from_prior) {
-      if (from_prior) {
-        list(mean = params@mean, prec = params@prec)
-      } else {
-        list(ref_dose = ref_dose, mean = params@mean, prec = params@prec)
+      ms <- list(mean = params@mean, prec = params@prec)
+      if (!from_prior) {
+        ms$ref_dose <- ref_dose
       }
+      ms
     },
     init = function() {
       list(theta = c(0, -20))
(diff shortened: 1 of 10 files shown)


R/Model-class.R
```diff
@@ -185,8 +197,12 @@ ModelLogNormal <- function(mean, cov, ref_dose = 1) {
...
       alpha1 <- exp(theta[2])
     },
-    modelspecs = function() {
-      list(ref_dose = ref_dose, mean = params@mean, prec = params@prec)
+    modelspecs = function(from_prior) {
+      if (from_prior) {
+        list(mean = params@mean, prec = params@prec)
+      } else {
+        list(ref_dose = ref_dose, mean = params@mean, prec = params@prec)
+      }
```
danielinteractive, line 205:
```suggestion
      ret <- list(mean = params@mean, prec = params@prec)
      if (!from_prior) {
        ret$ref_dose <- ref_dose
      } 
      ret
```

danielinteractive: this makes it explicit that we have additional stuff being added to the return list when it is not from prior.

tests/testthat/test-helpers_jags.R
```diff
@@ -92,10 +92,15 @@ test_that("h_jags_get_data works as expected", {
...
 
   result <- h_jags_get_data(model, data, FALSE)
+  expected_prior <- c(
+    h_slots(data, c("nObs", "y", "x")),
+    model@modelspecs(TRUE) # nolintr
+  )
   expected <- c(
     h_slots(data, c("nObs", "y", "x")),
-    model@modelspecs() # nolintr
+    model@modelspecs(FALSE) # nolintr
```
danielinteractive, line 101: in all these tests to make it better readable I would always say `from_prior = TRUE` etc. so with the argument name

danielinteractive (approved):
thx Wojciech, pls see inside some simplification/readability suggestions, can you please apply this to all relevant code changes? afterwards good to merge. thx

R/Model-class.R
wwojciech: Sure, I was even thinking about it but finally chose something different :)

PR #224, opened 2022-05-26
Closes #221

### LINKED ISSUE #221: Assertions for `mcmc` for `DualEndpoint` and `OneParExpNormalPrior`
The `mcmc` method for:
- [x] `DualEndpoint` classes should have assertion that `nGrid >= 2` or `nGrid >= 3`.
- [x] `OneParExpNormalPrior` class should have assertion that `length(skel_probs) == max(data@xLevel)` when `datamodel` is used (only).

wwojciech:
Closes #221

### DIFF
--- a/R/Model-class.R
+++ b/R/Model-class.R
@@ -1225,7 +1225,7 @@ DualEndpoint <- function(mean,
       z[i] ~ dnorm(meanZ[i], 1)
       y[i] ~ dinterval(z[i], 0)
 
-      # The conditional biomarker model. betaW defined in subclasses!
+      # The conditional biomarker model; betaW defined in subclasses!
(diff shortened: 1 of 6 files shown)


wwojciech:
Hi @danielinteractive! For `DualEndpoint` classes (`RW`, `beta`, `Emax`) I created here three 2 additional `mcmc` methods in order to assert the value of some object. Alternatively, I could create just one `mcmc` method for `DualEndpoint` class and do proper assertions for `RW`, `beta`, `Emax` subclasses using `if-else` clauses. I am not so sure which option is better. What do you think Daniel? Tx!

danielinteractive (approved):
thx Wojciech, good point that we need model specific assertions here, thx

wwojciech:
> thx Wojciech, good point that we need model specific assertions here, thx

Thank you @danielinteractive What do you think about the question I wrote above? Tx a lot!

PR #228, opened 2022-05-31
Closes #227

### LINKED ISSUE #227: Rules NextBestNCRM` class cleanup

To do:

- [ ] Revise documentation
  - [ ] Replace `##'` by `#'`
  - [ ] Add to `pkgdown.yml`
- [ ] Add unit tests
  - [ ] For classes:
  - [ ] For functions: as usual
- [ ] Refactor as needed
  - [ ] Remove `keywords` tags
  - [ ] For classes:

wwojciech:
Closes #227


R/Rules-class.R
```diff
@@ -78,74 +74,75 @@ NextBestMTD <- function(target, derive) {
...
 
-                     o$result()
-                 })
-validObject(.NextBestNCRM())
+#' `NextBestNCRM`
+#'
+#' @description `r lifecycle::badge("stable")`
+#'
+#' [`NextBestNCRM`] is the class for next best dose that finds the next dose in
+#' target interval.
```
danielinteractive, line 86:
```suggestion
#' [`NextBestNCRM`] is the class for next best dose that finds the next dose with high posterior probability to be in the
#' target toxicity interval.
```

tests/testthat/test-Rules-class.R
```diff
@@ -17,28 +17,22 @@ test_that("NextBestMTD object can be created with user constructor", {
   expect_valid(result, "NextBestMTD")
 })
 
-test_that("NextBestMTD throws the error for non valid target", {
```
danielinteractive, line 20: what about those NextBestMTD tests? why did you remove those?

tests/testthat/test-Rules-validity.R
```diff
@@ -29,6 +29,46 @@ test_that("v_next_best_mtd returns message for non-valid derive", {
...
+  # Changing `overdose` so that it is not an interval.
+  object@overdose <- 1.2
+  expect_equal(v_next_best_ncrm(object), "overdose has to be a probability range")
+
+  # Changing `overdose` so that the one bound is not a valid probability.
+  object@overdose <- c(0.4, 2)
+  expect_equal(v_next_best_ncrm(object), "overdose has to be a probability range")
+})
+
+test_that("h_next_best_ncrm returns message for non-valid max_overdose_prob", {
```
danielinteractive, line 61:
```suggestion
test_that("v_next_best_ncrm returns message for non-valid max_overdose_prob", {
```

danielinteractive (commented):
thx Wojciech, please see inside, and also add the test(s) for the `nextBest()` method for this class in `test-Rules-methods.R`, thanks!

tests/testthat/test-Rules-class.R
wwojciech: Yes, I removed intentionally, as I realised these were a duplicated unit tests for the functionality that is provided by class validation function, and for this we already have unit tests.

danielinteractive (commented):
Did you already add the test(s) for the nextBest() method?

PR #230, opened 2022-06-01
Closes #229

### LINKED ISSUE #229: move unit tests for`NextBestInfTheory` to correct R file.
`## Information Theory Approach` section with unit tests should be moved from `test-Model-methods.R` file to `test-Rules-methods.R`.

wwojciech:
Closes #229


wwojciech:
Seems that some two other tests dedicated for `NextBestInfTheory` class would need to be moved from `test-Rules-methods.R` file. But I am not sure at this stage, to where it should be moved: to `test-Rules-class.R` or `design` or `simulate` related unit test files (that do not yet exist).

danielinteractive (approved):

PR #234, opened 2022-06-01
Closes #231

### LINKED ISSUE #231: Rules `NextBestThreePlusThree` class cleanup

To do:

- [ ] Revise documentation
  - [ ] Replace `##'` by `#'`
  - [ ] Add to `pkgdown.yml`
- [ ] Add unit tests
  - [ ] For classes:
  - [ ] For functions: as usual
- [ ] Refactor as needed
  - [ ] Remove `keywords` tags
  - [ ] For classes:

wwojciech:
Closes #231


R/Rules-class.R
```diff
@@ -141,35 +141,39 @@ NextBestNCRM <- function(target,
...
+
+#' @rdname NextBestThreePlusThree-class
+#'
+#' @export
+#' @examples
+#' # Next best dose class object using the classical 3+3 design.
+#' my_next_best <- NextBestThreePlusThree()
+NextBestThreePlusThree <- function() {
+  .NextBestThreePlusThree()
 }
```
danielinteractive, line 174:
```suggestion
NextBestThreePlusThree <- .NextBestThreePlusThree
```
in this case even easier

danielinteractive (approved):

R/Rules-class.R
danielinteractive: ok no prob

danielinteractive (approved):

PR #235, opened 2022-06-02
Closes #232

### LINKED ISSUE #232: Rules `NextBestDualEndpoint` class cleanup

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #232


R/Rules-class.R
```diff
@@ -173,108 +173,108 @@ NextBestThreePlusThree <- function() {
...
-                     }
-                     o$check(is.probRange(object@overdose),
-                             "overdose has to be a probability range")
-                     o$check(is.probability(object@maxOverdoseProb),
-                             "maxOverdoseProb has to be a probability")
-                     o$check(is.probability(object@targetThresh),
-                             "targetThresh has to be a probability")
+#' `NextBestDualEndpoint`
+#'
+#' @description `r lifecycle::badge("stable")`
```
danielinteractive, line 182: actually we should set all DualEndpoint classes as `experimental`, that is more honest...

danielinteractive (approved):
thx Wojciech, just one side comment

R/Rules-class.R
wwojciech: Sure, Thanks @danielinteractive!

PR #250, opened 2022-06-13
Closes #249 (custom checkmate - probability)

### LINKED ISSUE #249: Propose new `checkmate` customised functions for `is.probability`
Propose new `checkmate` customised functions as an replacement for existing helpers for probability check, i.e.: `is.probability` and etc.

wwojciech:
Closes #249


R/checkmate.R
```diff
@@ -0,0 +1,176 @@
...
+#'
+#' @seealso [`assertions`] for more details.
+#'
+#' @export
+#' @examples
+#' x <- c(0, 0.2, 0.1, 0.3, 1)
+#' check_probability(x)
+#' check_probability(x, bounds_closed = FALSE)
+#' check_probability(x, bounds_closed = c(FALSE, TRUE))
+check_probability <- function(x, bounds_closed = TRUE, len = NULL, sorted = FALSE) {
```
danielinteractive, line 51: comment on naming: I would propose `check_probabilities` here and `check_probability` below

R/checkmate.R
```diff
@@ -0,0 +1,176 @@
...
+#' @seealso [`assertions`] for more details.
+#'
+#' @export
+#' @examples
+#' x <- c(0, 0.2)
+#' check_prob_range(x)
+#' check_prob_range(rev(x))
+#' check_prob_range(x, bounds_closed = FALSE)
+#' check_prob_range(x, bounds_closed = c(FALSE, TRUE))
+check_prob_range <- function(x, bounds_closed = TRUE) {
```
danielinteractive, line 160: then this can be `check_probability_ranges`

danielinteractive (commented):
thanks Wojciech, nice work! just would change the name of the functions slightly to make them more consistent

R/checkmate.R
wwojciech: Can it be `check_probability_range` are we check just one range?

R/checkmate.R
wwojciech: Thank you @danielinteractive, Sure, just one concern I have is that these names are very close to each other and hence easy to make a mistake. I was inspired by `check_integer` / `check_int` when proposed `check_probability` / `check_prob`. Let me know if you are sure about this change, and I will change it then.

R/checkmate.R
danielinteractive: ah yes

wwojciech:
@danielinteractive - do you think I should also add these 3 functions to `_pkgdown.yaml` file?

danielinteractive:
@wwojciech yes please

R/checkmate.R
wwojciech: Ok, so I will change it :)

danielinteractive (approved):

PR #139, opened 2022-02-27
New dose recommendation rule based on the loss function 

gunhanb:
closes #109 

New dose recommendation rule based on the loss function. This class is similar to NCRM class, the only difference is the addition of loss function. I have updated Rules-Class, Rules-Methods, and included examples and tests.

PR #257, opened 2022-06-14
Closes #109 (new branch based on latest main)

### LINKED ISSUE #109: A new class for the next best dose recommendation based on the loss function
**Summary**
This is a new rule for next best dose recommendation. Next dose derivation is based on minimizing loss function while controlling the probability of excessive/unacceptable toxicity. The loss function based recommendation was described in Neuenschwander et al (2008). 
[Rules_NCRM_loss.txt](https://github.com/Roche/crmPack/files/8068215/Rules_NCRM_loss.txt)

gunhanb:
Closes #109; New branch based on latest main. It contains initial new code for Rules - but it should be further completed.


danielinteractive:
@gunhanb @wwojciech can you please work together to get checks fixed and then merge? thanks!

gunhanb:
Hi Daniel,

This is done with @wwojciech actually. I am in contact with him to resolve this.

R/Rules-class.R
```diff
@@ -142,6 +142,94 @@ NextBestNCRM <- function(target,
...
+#'
+#' @aliases NextBestNCRMLoss
+#' @export
+#'
+.NextBestNCRMLoss <- setClass(
+  Class = "NextBestNCRMLoss",
+  slots = c(
+    target_int = "numeric",
+    overdose_int = "numeric",
+    unacceptable_int = "numeric",
```
wwojciech, line 186: should not be an 'integer' ?

R/Rules-validity.R
```diff
@@ -48,6 +48,35 @@ v_next_best_ncrm <- function(object) {
...
+    test_probability(object@max_overdose_prob, bounds_closed = FALSE),
+    "max_overdose_prob must be a probability value from (0, 1) interval"
+  )
+  v$check(
+    all(object@losses >= 0),
+    "losses has to be a numeric vector with all elements non-negative"
+  )
+  v$result()
+}
+
```
wwojciech, line 79: I think this validation should be improved. For instance, 

1. what if condition in `if` clause is `FALSE`?
2. for `losses` please use `checkmate`
3. `overdose_int` - left bound in the interval is opened right? same for `unacceptable_int`.
4. lengths of `max_overdose_prob` and `losses` should also be verified.

wwojciech (changes requested):
Looks good but some important changes are required.

R/Rules-validity.R
wwojciech: Sure; in particular `checkmate` provides a set of testing function `test_*` that we use to validate the objects.
For example 
```
test_numeric(object@weights, lower = 0, finite = TRUE, any.missing = FALSE, len = 2)
```
tests wether  `object@weights` is a numeric vector of length 2 without any missing values, with all values finite and greater than or equal to 0.

See more examples e.g. in `Model-validity.R` file.

danielinteractive:
@wwojciech @gunhanb how about you have another 1:1 session to finish this? so instead of review comments finish it together. just would be good to merge this. thanks

wwojciech:
> @wwojciech @gunhanb how about you have another 1:1 session to finish this? so instead of review comments finish it together. just would be good to merge this. thanks

Sure @danielinteractive , @gunhanb let me know please when it is convenient for you to have another 1:1 session. Tx!

wwojciech:
Thanks @gunhanb I see it now fails at some `linter` checking. Would you like me to correct these ones or you correct it by yourself?

gunhanb:
> Thanks @gunhanb I see it now fails at some `linter` checking. Would you like me to correct these ones or you correct it by yourself?

Would be great if you can finalize it? Sorry for the additional work!

wwojciech:
> > Thanks @gunhanb I see it now fails at some `linter` checking. Would you like me to correct these ones or you correct it by yourself?
> 
> Would be great if you can finalize it? Sorry for the additional work!

@gunhanb  - sure, no problem. So I will finish all the rest (including some cleanup and etc, if any) and then accept it.

wwojciech (dismissed):

wwojciech:
Hi @danielinteractive , I would merge this PR as it is now (without `nextBest`) to not loose its current good sync with the `main` that progresses quite fast. Then, `nextBest` can be created in a separate issue. What do you think?

wwojciech:
> @wwojciech can you maybe check if we lost the method somewhere in the process? e.g. did the old PR have this? I would not like to merge like it is now, feels a bit weird

Your right @danielinteractive, it is in the old PR ( https://github.com/Roche/crmPack/pull/139 ). Thanks for the hint! Unfortunately, it has no unit tests at all, and it needs to be cleaned up....

wwojciech:
I copied-pasted `nextBest` method for `NCRMLoss` but it fails on the examples provided.

@gunhanb - are you sure this below is correct?
```
#' @slot losses (`numeric`)\cr a vector specifying the loss function. If the
#'   `unacceptable` is provided, the vector length must be \eqn{4}, otherwise
#'   \eqn{3}.
```
The `nextBest` method you wrote fails, when specifying `losses` with this rule. I think it should be the opposite, i.e. if the `unacceptable` is provided, the vector length must be of length `3`, and otherwise it must be `4`. Can you please clarify this and check the implementation?

wwojciech:
> > I copied-pasted `nextBest` method for `NCRMLoss` but it fails on the examples provided.
> > @gunhanb - are you sure this below is correct?
> > ```
> > #' @slot losses (`numeric`)\cr a vector specifying the loss function. If the
> > #'   `unacceptable` is provided, the vector length must be \eqn{4}, otherwise
> > #'   \eqn{3}.
> > ```
> > 
> > The `nextBest` method you wrote fails, when specifying `losses` with this rule. I think it should be the opposite, i.e. if the `unacceptable` is provided, the vector length must be of length `3`, and otherwise it must be `4`. Or maybe this rule is OK and the `nextBest` method implementation needs to be corrected. Can you please clarify this and check the implementation?
> 
> Yes, you are absolutely right. It is a typo, it should be the other way around. When `unacceptable` is provided, the vector length must be of length `4`, and otherwise it must be `3`. Sorry for the mistake!

@gunhanb - sorry I am a bit confused, it is already like that (when `unacceptable` is provided, the vector length is of 
 length `4`, and otherwise it is be `3`). But you write in the same sentence that I am right and it should be the opposite?

gunhanb:
> 

Ahh I confused myself! The original documentation is correct, that is when unacceptable is provided, the vector length is of
length 4, and otherwise it is be 3.

wwojciech:
> > 
> 
> Ahh I confused myself! The original documentation is correct, that is when unacceptable is provided, the vector length is of length 4, and otherwise it is be 3.

Got it, Tx! So then, there is some serious bug in the `nextBest-NextBestNCRMLoss` method implementation, as it fails for the valid example you provided in `Rules-method-nextBest-NextBestNCRMLoss.R` file.
The failure is: `Error in nextBest@losses %*% t(probMat) : non-conformable arguments`, so I would say that it expects that length of `losses` is equal `3`, when `unacceptable` is provided. Can you pl. check this? Tx!

danielinteractive:
@gunhanb @wwojciech maybe again would be more efficient to schedule  a short meeting to iron this out :-)

R/Rules-methods.R
```diff
@@ -284,6 +284,219 @@ setMethod(
...
+
+    # Matrix with samples from the dose-tox curve at the dose grid points.
+    samples_prob <- sapply(data@doseGrid, prob, model = model, samples = samples)
+
+    prob_underdosing <- colMeans(samples_prob < nextBest@target[1])
+    prob_target <- colMeans(
+      (samples_prob >= nextBest@target[1]) & (samples_prob <= nextBest@target[2])
+    )
+    prob_overdose <- colMeans(
+      (samples_prob > nextBest@overdose[1]) & (samples_prob <= nextBest@overdose[2])
```
wwojciech, line 316: But the `nextBest@overdose`, `nextBest@target` are never set when creating `nextBest` class object with `NextBestNCRMLoss()` user constructor. Something is wrong here.

wwojciech (changes requested):
Hi @gunhanb - I think that 3 slots that I pointed out in the definition of `NextBestNCRMLoss` are obsolete. Then the implementation of `nextBest` methods needs to be adjusted accordingly. Feel free to reach out in case of any additional questions or doubts. Tx!

wwojciech (approved):
Hi @gunhanb, It looks good now, Thanks a lot!

PR #258, opened 2022-06-15
Closes #254

### LINKED ISSUE #254: Rules `NextBestInfTheory` class cleanup

To do:

- [ ] Revise documentation
  - [ ] Replace `##'` by `#'`
  - [ ] Add to `pkgdown.yml`
- [ ] Add unit tests
  - [ ] For classes:
  - [ ] For functions: as usual
- [ ] Refactor as needed
  - [ ] Remove `keywords` tags
  - [ ] For classes:

wwojciech:
Closes #254


wwojciech:
Hi @danielinteractive - I had to slightly modify/rearrange few existing unit tests for `Rules` classes. I remember it would be better to start a new PR for such a separate modifications, but I initially thought it would be very small - but ended up with a bit more than I expected :(. I do apologise for the inconveniences caused.

danielinteractive (approved):
thx Wojciech, looks good to me! maybe try pinging @pc494 in case he can have a quick look too? (re: default values e.g.)

wwojciech:
Tx @danielinteractive 
I actually checked it a bit with this document - https://onlinelibrary.wiley.com/doi/10.1002/sim.8450, but sure I will let Phillip know about it.

wwojciech:
Hi Phillip (@pc494) !
Within this issue and PR I refactored `NextBestInfTheory` Rules class you created. Daniel suggested it would be good if you have a time to look at it after my refactoring to ensure default values and validation function I created makes sense. Tx!

wwojciech:
Thanks a lot Phillipe @pc494 !

PR #262, opened 2022-06-20
Closes #259

### LINKED ISSUE #259: `Rules` cleanup: `nextBest` method for `NextBestMTD` signature

To do:

- [ ] Revise documentation
  - [ ] Replace `##'` by `#'`
  - [ ] Add to `pkgdown.yml`
- [ ] Add unit tests
  - [ ] For classes:
  - [ ] For functions: as usual
- [ ] Refactor as needed
  - [ ] Remove `keywords` tags
  - [ ] For classes:

wwojciech:
Closes #259


R/Rules-methods.R
```diff
@@ -4,127 +4,133 @@
...
+#'
+#' A function that computes the recommended next best dose based on the
+#' corresponding rule `nextBest`, the posterior `samples` from the `model` and
+#' the underlying `data`.
+#'
+#' @param nextBest (`NextBest`)\cr the rule for the next best dose.
+#' @param doselimit (`number`)\cr the maximum allowed next dose. If it an empty
+#'   vector, then no dose limit will be applied in the course of dose
+#'   recommendation calculation, and a corresponding warning is given.
+#' @param samples (`Samples`)\cr a samples.
```
danielinteractive, line 23:
```suggestion
#' @param samples (`Samples`)\cr posterior samples from `model` parameters given `data`.
```

R/Rules-methods.R
```diff
@@ -4,127 +4,133 @@
...
+#' @param model (`Model`)\cr model input.
+#' @param data (`Data`)\cr data input.
```
danielinteractive, line 25:
```suggestion
#' @param model (`Model`)\cr used model.
#' @param data (`Data`)\cr used data.
```
(just to allude to that this was used to generate the samples)

R/Rules-methods.R
```diff
@@ -4,127 +4,133 @@
...
+#' @param ... additional arguments without method dispatch.
+#'
+#' @return a list with the next best dose recommendation  (element named `value`)
```
danielinteractive, line 28:
```suggestion
#' @return A list with the next best dose recommendation  (element named `value`)
```

R/Rules-methods.R
```diff
@@ -4,127 +4,133 @@
...
+    mtd_samples <- dose(x = nextBest@target, model, samples)
+    mtd_estimate <- nextBest@derive(mtd_samples = mtd_samples)
+    # Round to the next possible grid point.
+    min_dist_ind <- which.min(abs(doseGrid_leq_limit - mtd_estimate))
+    ret <- doseGrid_leq_limit[min_dist_ind]
+
+    # Create a plot.
+    p <- ggplot(
+      data = data.frame(x = mtd_samples),
+      aes(x = x),
```
danielinteractive, line 96: I guess we need
```suggestion
      aes(x = .data$x),
```
to avoid `R CMD check` note, also for the other aesthetics. See e.g. https://github.com/insightsengineering/hermes/blob/ce68571f3ccac0fa17945587fb587f911f604c81/R/draw_barplot.R#L86 as example

tests/testthat/test-Rules-methods.R
```diff
@@ -1,5 +1,46 @@
...
+  nb_mtd <- NextBestMTD(
+    target = 0.33,
+    derive = function(mtd_samples) {
+      quantile(mtd_samples, probs = 0.25)
+    }
+  )
+
+  result <- nextBest(nb_mtd, 90, samples, model, data)
+  expect_identical(result$value, 75)
+  vdiffr::expect_doppelganger("Plot of nextBest-NextBestMTD", result$plot)
```
danielinteractive, line 20: now looking at the plot - how about rotating the labels of the vertical lines 90 degrees such that they overlap less? just an idea to make it a bit better readable

danielinteractive (commented):
Thank you Wojciech, good start of the `nextBest` business here! See few comments inside

danielinteractive (approved):

PR #271, opened 2022-06-30
Closes #270

### LINKED ISSUE #270: Rules methods cleanup: `NextBestThreePlusThree`

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #270


R/Rules-methods.R
```diff
@@ -284,77 +284,59 @@ setMethod(
...
-                  if(nextLevel == 0)
-                  {
-                      TRUE
-                  } else {
-                      nPatients[nextLevel] > 3
-                  }
+    next_level <- last_level + if (DLT_rate_last_level < 1 / 3) {
+      # Escalate it, unless this is the highest level or the higher level
+      # already tried (and it was not safe).
+      ifelse((last_level == data@nGrid) || (nPatients[last_level + 1] > 0), 0, 1)
```
wwojciech, line 318: Hi @danielinteractive , I have one doubt to the comment that is in the brackets in the line 317, that is I think it if not true for the code in the line 318, as it does not check whether the higher level that was already tried is safe or not.

danielinteractive: Hi @wwojciech well implicitly we know that it was not safe - because only then we would ever go down again

wwojciech: Tx @danielinteractive , so it means that we should assume that such data is not possible?
```
data <- Data(
  x = c(25, 25, 100, 75, 75, 75, 75),
  y = c(0L, 0L, 0L, 0L, 0L, 0L, 1L),
  doseGrid = c(25, 50, 75, 100, 125)
)
```
Because, here the last dose it `75` and the higher `100` was already tried before and it did not have any DLT...

danielinteractive: such data won't be possible I think. feel free to modify the comment though if you think too many assumptions behind it

danielinteractive: I would not change the algorithm though.

R/Rules-methods.R
```diff
@@ -284,81 +284,58 @@ setMethod(
...
-                  } else {
-                      nPatients[nextLevel] > 3
-                  }
-
-              ## return value and plot
-              return(list(value=
-                          if(nextLevel == 0) NA else data@doseGrid[nextLevel],
-                          stopHere=stopHere))
-          })
+    next_level <- last_level + if (DLT_rate_last_level < 1 / 3) {
```
danielinteractive, line 315: I would first assign the result of the if/else to a variable, e.g. `level_change` and then add that to `last_level`. otherwise too hard to read.

R/Rules-methods.R
```diff
@@ -284,81 +284,58 @@ setMethod(
...
+      # Escalate it, unless this is the highest level or the higher already tried.
```
danielinteractive, line 316:
```suggestion
      # Escalate it, unless this is the highest level or the higher dose was already tried.
```

danielinteractive (approved):
thanks Wojciech, looks good!

PR #273, opened 2022-07-08
Closes #272

### LINKED ISSUE #272: Rules methods cleanup: `NextBestDualEndpoint`

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #272


R/Rules-methods.R
```diff
@@ -305,28 +305,25 @@ setMethod("nextBest",
...
+    prob_overdose <- colMeans(h_in_range(prob_samples, nextBest@overdose, c(FALSE, TRUE)))
+    prob_mean <- colMeans(prob_samples)
+    prob_sd <- apply(prob_samples, 2, sd)
 
-    prob_underdosing <- colMeans(samples_prob < nextBest@target[1])
-    prob_target <- colMeans(h_in_range(samples_prob, nextBest@target))
-    prob_overdose <- colMeans(h_in_range(samples_prob, nextBest@overdose, c(FALSE, TRUE)))
-    prob_mean <- colMeans(samples_prob)
-    prob_sd <- apply(samples_prob, 2, sd)
+    is_unacceptable_specified <- any(nextBest@unacceptable != c(1, 1))
```
danielinteractive, line 317: so this is a fix for the `NextBestNCRMLoss` method right?

wwojciech:
HI @danielinteractive It seems that ploting code for some `nextBest` methods is quite similar but also long. Do you think it is reasonable to extract the ploting code from `nextBest` methods into separate function. Not necessary helpers, but maybe regular generics/methods like `nextBext-plot`? Whad do you think? Tx a lot!

R/Rules-methods.R
```diff
@@ -526,259 +521,136 @@ setMethod(
...
+    data = "Data"
+  ),
+  definition = function(nextBest, doselimit, samples, model, data, ...) {
+    doselimit <- ifelse(missing(doselimit) || length(doselimit) == 0, Inf, doselimit)
 
-        ret <- data@doseGrid[dosesOK][doseLevel]
+    # Biomarker samples at the dose grid points.
+    biom_samples <- samples@data$betaW
+
+    if (nextBest@target_relative) {
```
danielinteractive, line 551: I guess the result of this if/else blocks is `prob_target`, therefore would recommend to use the pattern
```r
prob_target <- if () else .. etc.
```

R/Rules-methods.R
```diff
@@ -526,259 +521,136 @@ setMethod(
...
-          colour = "darkgreen",
-          fill = "darkgreen"
-        ) +
-        xlab("Dose") +
-        ylab(paste("Target probability [%]")) +
-        ylim(c(0, 100))
+    # Now, compute probabilities to be in overdose tox interval, then flag
+    # eligible grid doses after accounting for maximum possible dose and discarding overdoses.
+    prob_samples <- sapply(data@doseGrid, prob, model = model, samples = samples)
+    prob_overdose <- colMeans(h_in_range(prob_samples, nextBest@overdose, c(FALSE, TRUE)))
```
danielinteractive, line 574: I think it would be a bit better readable if you use the argument name for the last logical argument

R/Rules-methods.R
```diff
@@ -526,259 +521,136 @@ setMethod(
...
+    is_dose_eligible <- data@doseGrid <= doselimit & prob_overdose < nextBest@max_overdose_prob
```
danielinteractive, line 575:
```suggestion
    is_dose_eligible <- (data@doseGrid <= doselimit) & (prob_overdose < nextBest@max_overdose_prob)
```

R/Rules-methods.R
```diff
@@ -526,259 +521,136 @@ setMethod(
...
-      if (length(doselimit)) {
-        plot1 <- plot1 +
-          geom_vline(
-            xintercept = doselimit,
-            lwd = 1.1,
-            lty = 2,
-            colour = "black"
-          )
-      }
+    next_best <- if (any(is_dose_eligible)) {
```
danielinteractive, line 581: yeah this is the pattern, nice :-)

R/crmPack-package.R
```diff
@@ -33,7 +33,7 @@
 #' @importFrom graphics plot hist legend lines matlines matplot
 #' @importFrom stats binomial coef cov2cor gaussian glm lm median model.matrix
 #' optim pgamma plogis pnorm qgamma qlogis qnorm quantile rbinom rgamma approxfun
-#' rnorm runif uniroot var vcov step mad pbeta dbeta dgamma setNames
+#' rnorm runif uniroot var vcov step mad pbeta dbeta dgamma setNames sd
```
danielinteractive, line 36: instead of this (I thought I mentioned this before but not sure) would be better to just import a single function and always use `stats::fun` for any `fun`.

danielinteractive (commented):
thanks a lot Wojciech! only few comments inside

R/crmPack-package.R
wwojciech: Yap, you mentioned about this - but I was not sure which option was the correct one, and I chose wrong one :( sorry :).

R/Rules-methods.R
wwojciech: Yap, it is a fix (that was applied in the Burak's branch and merged into main. so I just merged an updated main into this #273 branch) + few cosmetic changes after the fix, which I done in this branch.

R/Rules-methods.R
wwojciech: Yap, absolutely, thanks I just missed that.

R/Rules-methods.R
wwojciech: Sure, the line will be a bit longer but I guess this is ok.

R/Rules-methods.R
danielinteractive: ah ok good. thx

danielinteractive (approved):

PR #278, opened 2022-07-13
Closes #277

### LINKED ISSUE #277: Rules methods cleanup: NextBestInfTheory

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #277

### DIFF
--- a/R/helpers_rules.R
+++ b/R/helpers_rules.R
@@ -6,14 +6,19 @@
 #' divergence as given by equation in (7) in the reference at
 #' https://doi.org/10.1002/sim.8450.
 #'
-#' @param prob (`numeric`)\cr probability of a DLT occurring.
-#' @param target (`numeric `)\cr target probability of a DLT.
+#' @param prob (`numeric`)\cr vector or matrix with probabilities of a DLT occurring.
+#' @param target (`number `)\cr single target probability of a DLT.
 #' @param asymmetry (`number`)\cr describes the rate of penalization
 #'   for overly toxic does, range 0 to 2.
 #'
 #' @export
 #' @examples
-#' h_info_theory_dist(c(0.5, 0.2), c(0.5, 0.1), 1.2)
+#' h_info_theory_dist(c(0.5, 0.2), 0.4, 1.2)
 h_info_theory_dist <- function(prob, target, asymmetry) {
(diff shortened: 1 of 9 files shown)


tests/testthat/test-helpers_rules.R
```diff
@@ -1,11 +1,38 @@
...
+    h_info_theory_dist(0.2, c(0.4, 0.5), 1.2),
+    "Assertion on 'target' failed: Must have length 1."
+  )
+  expect_error(
+    h_info_theory_dist(0.2, 0.4, c(1.2, 1.6)),
+    "Assertion on 'asymmetry' failed: Must have length 1."
+  )
+})
+
+test_that("h_info_theory_dist trhows the error for wrong asymmetry", {
```
danielinteractive, line 33:
```suggestion
test_that("h_info_theory_dist throws the error for wrong asymmetry", {
```

R/helpers_rules.R
```diff
@@ -6,14 +6,19 @@
...
-#' @param target (`numeric `)\cr target probability of a DLT.
+#' @param prob (`numeric`)\cr vector or matrix with probabilities of a DLT occurring.
+#' @param target (`number `)\cr single target probability of a DLT.
 #' @param asymmetry (`number`)\cr describes the rate of penalization
 #'   for overly toxic does, range 0 to 2.
 #'
 #' @export
 #' @examples
-#' h_info_theory_dist(c(0.5, 0.2), c(0.5, 0.1), 1.2)
+#' h_info_theory_dist(c(0.5, 0.2), 0.4, 1.2)
```
danielinteractive, line 16: Makes sense, I guess the example worked before too because of vector calculations

danielinteractive (approved):
nice work, thanks @wwojciech !

R/helpers_rules.R
wwojciech: Yap, but I think at the moment we need to to be more secure rather than some extra flexibility, never used :)

PR #281, opened 2022-07-18
Closes #279

### LINKED ISSUE #279: Rules methods cleanup: `NextBestTDsamples`

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #279


R/Rules-methods.R
```diff
@@ -717,6 +717,119 @@ setMethod(
...
+    # closest and below the target_in_trial_est (or target_trial_end_est, respectively).
+    # h_find_interval assumes that elements in doses_eligible are strictly increasing.
+    next_best_level <- h_find_interval(target_in_trial_est, doses_eligible)
+    next_best <- doses_eligible[next_best_level]
+
+    next_best_level1 <- h_find_interval(target_trial_end_est, doses_eligible)
+    next_best1 <- doses_eligible[next_best_level1]
+
+    # 95% credibility interval.
+    CITDEOT <- as.numeric(quantile(target_trial_end_samples, probs = c(0.025, 0.975)))
```
danielinteractive, line 764: please use snake_case instead of all caps for variable names

R/Rules-methods.R
```diff
@@ -717,6 +717,119 @@ setMethod(
...
+    TDEOT_ratio <- CITDEOT[2] / CITDEOT[1]
+
+    # Build plot.
```
danielinteractive, line 767: the function is very long, suggest to consider factoring out plot in separate helper function

danielinteractive (commented):
thx Wojciech, nice work, please see few comments inside

R/Rules-methods.R
wwojciech: Yap, I agree. This was actually my question when started to work on `nextBest` methods - https://github.com/Roche/crmPack/pull/273#issuecomment-1178861547

danielinteractive: Hm I would not do a separate file necessarily, just putting the helper function above the method could work. And better do it incrementally as you hit those long methods instead of doing it all together at once

wwojciech: I see. Tx @danielinteractive. However, we already have `helpers_rules.R` file for this purpose, for example it contains already some helpers like that for `nextBest` methods? 

Also - since this new helper function which creates a plot is to be used only by this particular `nextBest` methods do you think we need to add another set of unit test just for this helper? or it is ok if we test it within `nextBest` as it is already?

danielinteractive: In any case we need to test the plot results, so you can go with the simpler way, either unit test for the plot helper or unit test for the method.

PR #284, opened 2022-07-21
Closes #282

### LINKED ISSUE #282: Rules methods cleanup: `NextBestTD`

To do:

- [ ] Revise documentation
  - [ ] Replace `##'` by `#'`
  - [ ] Add to `pkgdown.yml`
- [ ] Add unit tests
  - [ ] For classes:
  - [ ] For functions: as usual
- [ ] Refactor as needed
  - [ ] Remove `keywords` tags
  - [ ] For classes:

wwojciech:
Closes #282


R/Rules-methods.R
```diff
@@ -790,6 +790,106 @@ setMethod(
...
+    # at level closest and below the target dose estimate.
+    # h_find_interval assumes that elements in doses_eligible are strictly increasing.
+    next_dose_lev_drt <- h_find_interval(dose_target_drt, doses_eligible)
+    next_dose_drt <- doses_eligible[next_dose_lev_drt]
+
+    next_dose_lev_eot <- h_find_interval(dose_target_eot, doses_eligible)
+    next_dose_eot <- doses_eligible[next_dose_lev_eot]
+
+    # Find the variance of the log of the dose_target_eot.
+    M <- matrix(
```
danielinteractive, line 843: Maybe avoid too short variable name (and upper case) here

danielinteractive (approved):

PR #301, opened 2022-08-18
Closes #280

### LINKED ISSUE #280: Remove `Model` class.
Because `Model` will be removed.

wwojciech:
Closes #280


danielinteractive (commented):
thanks Wojciech looks good however maybe then do at the same time the removal of `Model`?

wwojciech:
> thanks Wojciech looks good however maybe then do at the same time the removal of `Model`?

Thank you Daniel!, I think that there are some blocking places (before cleanup), where `the Model` might still be needed. But let me check this, maybe we can do it now for all and remove `Model` completely.

danielinteractive (approved):

PR #305, opened 2022-08-30
108 stopping rule assessed at specific dose

### LINKED ISSUE #108: Stopping rules assessed for specific dose and not just for the nextbest dose
To do:

- [x] See design doc https://github.com/Roche/crmPack/blob/main/design/stop-specific-dose.Rmd
- [x] Implement `v_stop_specific_dose()` validation function
- [x] Implement `StopSpecificDose` class
- [x] Implement `StopSpecificDose()` user constructor
- [x] Implement `stopTrial()` method for `StopSpecificDose`

Background: 

**Summary**

When defining a simulation design, we would like to have the option to  assess if Stopping rules are met for specific doses of the planned dose grid and not just for the nextBest dose.

For example assume we have a data set-up like below:

     Data(doseGrid = c(1.5, 2.5, 3.5, 4.5, 6, 7 ))

and that we want to implement a stopping rule as part of our dose escalation design simulations:
->Assuming a DLT rate of 30 %, we set=up our stopping rules to constantly check during each iteration, if the following conditions are met:

   stopping_rule_1 -> Minimum dose tested (1.5MBq) is toxic (DLT<=0.3) with probability T1 < 20%
   stopping_rule_2 -> Maximum possible dose (7.0MBq) is safe (DLT>0.3) with probability TN > 80% and 7.0MBq administered.

We could could rush and try set this up using stopping rules already available in the crmPACK for example and
only define a new class for stopping rule StoppingLowestDose. 
This could look like the below setting:

stopping_rule_1 <- StoppingLowestDose() & 
                                StoppingPatientsNearDose(nPatients=3, percentage=0) &
                                StoppingTargetProb(target=c(0.3, 1), prob=0.8)

stopping_rule_2 <- StoppingHighestDose() &
                               StoppingPatientsNearDose(nPatients=3,percentage=0) &
                               StoppingTargetProb(target=c(0, 0.3),prob=0.8)

But this would NOT really work correctly, because all the stopping rules during the design simulation process, 
are currently assessed at the nextbest dose  and not at a predefined dose of the planned dose grid.

A possible solution could be:
  1. An extension of the currently available stopping rules with an extra user input parameter(s) (slot(s)),
         where the dose of interest is defined OR else the default nextbest dose is used?
  2. An extension to the Design or Simulation object?
  3. Any other suggestion?

dimitriskontosClinBAY:
# Pull Request

closes #108


dimitriskontosClinBAY:
Do we need to create test-Rules for 
- method?
- class ?
- validity?

danielinteractive (commented):
Thanks @dimitriskontosClinBAY ! Yes can you please add all the unit tests ? thanks

examples/Rules-class-CohortSizeConst.R
```diff
@@ -1,4 +1,3 @@
 
 # This is to have along the study a constant cohort size of 3
-mySize <- CohortSizeConst(size=3)
-
+my_Size <- CohortSizeConst(size = 3)
```
danielinteractive, line 3:
```suggestion
my_size <- CohortSizeConst(size = 3)
```

examples/Rules-class-CohortSizeMax.R
```diff
@@ -1,17 +1,21 @@
 
 # Rule for having cohort of size 1 for doses <30
 #      and having cohort of size 3 for doses >=30
-mySize1 <- CohortSizeRange(intervals = c(0, 10),
-                           cohortSize = c(1, 3))
+mySize1 <- CohortSizeRange(
```
danielinteractive, line 4:
```suggestion
my_size_1 <- CohortSizeRange(
```

danielinteractive (commented):
just left some pointers to fix linter problems

R/Rules-class.R
```diff
@@ -1776,6 +1777,53 @@ StoppingAny <- function(stopList)
...
+## class ----
+
+#' `StoppingSpecificDose`
+#'
+#' @description `r lifecycle::badge("experimental")`
+#'
+#' [`StoppingSpecificDose`] is the class for testing a stopping rule at specific dose
+#'   of the dose grid and not at the next best dose
+#'
+#' @slot rule (`rule`)\cr a stopping rule available in this package.
```
wwojciech, line 1793:
```suggestion
#' @slot rule (`Stopping`)\cr a stopping rule available in this package.
```

R/Rules-class.R
```diff
@@ -1776,6 +1777,53 @@ StoppingAny <- function(stopList)
...
+  ),
+  contains = "Stopping",
+  validity = v_stopping_specific_dose
+)
+
+## constructor ----
+
+#' @rdname StoppingSpecificDose-class
+#'
+#' @param rule (`stopping rule`)\cr see slot definition.
```
wwojciech, line 1813:
```suggestion
#' @param rule (`Stopping`)\cr see slot definition.
```

R/Rules-validity.R
```diff
@@ -436,3 +436,14 @@ v_stopping_mtd_cv <- function(object) {
...
 #'  object contains valid probability target, threshold and shape parameters.
 v_stopping_lowest_dose_hsr_beta <- v_increments_hsr_beta # nolintr
+
+#' @describeIn v_stopping validates that the [`StoppingSpecificDose`] object
+#'   contains a stopping `rule` and a valid `dose`.
+v_stopping_specific_dose <- function(object) {
+  v <- Validate()
+  v$check(
+    test_number(object@dose, lower = 0, finite = TRUE),
+    "dose needs to be a single non zero finite number"
```
wwojciech, line 446: Is `0` allowed or not? If not then, `lower = .Machine$double.xmin` also maybe better to write: "dose needs to be a positive number"

R/Model-validity.R
```diff
@@ -49,19 +49,19 @@ v_model_logistic_kadane <- function(object) {
 v_model_logistic_kadane_beta_gamma <- function(object) { # nolintr
   v <- Validate()
   v$check(
-    test_number(object@alpha, lower = .Machine$double.xmin, finite = TRUE),
+    is.scalar(object@alpha) & is.numeric(object@alpha) && object@alpha > 0,
     "Beta distribution shape parameter alpha must be a positive scalar"
```
wwojciech, line 53: Hmm, why do we have this change here? I think `test_number` is better than `is.scalar and etc.` I guess this results from some merging issues.

R/Rules-methods.R
```diff
@@ -1233,15 +1237,17 @@ setMethod("maxDose",
...
-##' @example examples/Rules-method-maxDose-IncrementsRelativeDLT.R
+#' @rdname maxDose
+#'
+#' @description Increment rule to determine the maximum possible next dose
+#'   based on relative increments determined by DLTs so far.
+#'
+#' @aliases maxDose-IncrementsRelativeDLT
+#' @example examples/Rules-method-maxDose-IncrementsRelativeDLT.R
+#' @export
+#'
```
wwojciech, line 1250: Hmm, this seems again to be some merging issues.

wwojciech (changes requested):

dimitriskontosClinBAY:
Reviewer noticed many issues and problems (some wrong old code and etc) due to merging a main branch into this branch. It becomes risky to merge this current branch into main now. Was suggested to close this one without merging, and create a new, fresh branch from main to avoid any problems.

PR #307, opened 2022-08-31
Closes #304

### LINKED ISSUE #304: `NextBestTDsamples` Rules class cleanup.

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #304


R/Rules-class.R
```diff
@@ -448,6 +448,77 @@ NextBestInfTheory <- function(target, asymmetry) {
...
+#' for allocation during the trial and the second for final recommendation at
+#' the end of a trial. Hence, there are two target probabilities of the
+#' occurrence of a DLT that must be specified: target probability to be used
+#' during the trial and target probability to be used at the end of the trial.
+#'
+#' @slot prob_target_drt (`proportion`)\cr the target probability of the
+#'   occurrence of a DLT to be used during the trial.
+#' @slot prob_target_eot (`proportion`)\cr the target probability of the
+#'   occurrence of a DLT to be used at the end of the trial.
+#' @slot derive (`function`)\cr the function which derives, based on a vector of
```
danielinteractive, line 470:
```suggestion
#' @slot derive (`function`)\cr derives, based on a vector of
```

R/Rules-class.R
```diff
@@ -448,6 +448,77 @@ NextBestInfTheory <- function(target, asymmetry) {
...
+  validity = v_next_best_td_samples
+)
+
+## constructor ----
+
+#' @rdname NextBestTDsamples-class
+#'
+#' @param prob_target_drt (`proportion`)\cr the target probability of the
+#'   occurrence of a DLT to be used during the trial.
+#' @param prob_target_eot (`proportion`)\cr the target probability of the
```
danielinteractive, line 503: do we always repeat all the documentation in the constructor? maybe we could just say generically `see slot definition` ?

danielinteractive (approved):
thanks Wojciech, I guess the method is going to be cleaned up and tested in subsequent step?

R/Rules-class.R
wwojciech: Yes, we do, unfortunately :( I was also thinking about this today, I agree, it would be better to use `see slot definition`.

R/Rules-class.R
wwojciech: The `nextBest` method for `NextBestTDsamples` is already cleaned up and tested. 
I apologise, I am not sure if I understand this particular change request. The `derive` slot is a function. Should I change the description somehow?

R/Rules-class.R
wwojciech: I will clean this for `Rules`, and for separates files that were already cleaned up, I will create a new issue.

wwojciech:
> thanks Wojciech, I guess the method is going to be cleaned up and tested in subsequent step?

The `nextBest` method for `NextBestTDsamples` is already cleaned up and tested.

PR #309, opened 2022-08-31
Closes #306

### LINKED ISSUE #306: Modify validation for `derive` slot in `NextBestMTD` rules class.
The validation for `derive` function should be changed:

- [x] The name of the argument should not be validated, because it does not matter in fact.
- [x] It should be verified that the function has one parametr only (see recent changes in `NextBestTDsamples` as an example).
- [x] It should be verified that the function accepts a numerical vector parameter and returns a number. (see recent changes in `NextBestTDsamples` as an example).

wwojciech:
Closes #306

### DIFF
--- a/R/Rules-validity.R
+++ b/R/Rules-validity.R
@@ -22,8 +22,12 @@ v_next_best_mtd <- function(object) {
     "target must be a probability value from (0, 1) interval"
   )
   v$check(
-    formalArgs(object@derive) == "mtd_samples",
-    "derive must have a single argument 'mtd_samples'"
+    length(formalArgs(object@derive)) == 1L,
+    "derive must have a single argument"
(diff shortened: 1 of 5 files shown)


R/Rules-validity.R
```diff
@@ -22,8 +22,12 @@ v_next_best_mtd <- function(object) {
     "target must be a probability value from (0, 1) interval"
   )
   v$check(
-    formalArgs(object@derive) == "mtd_samples",
-    "derive must have a single argument 'mtd_samples'"
+    length(formalArgs(object@derive)) == 1L,
```
danielinteractive, line 25: you can use `test_function` from `checkmate` to simplify this

danielinteractive (approved):

PR #339, opened 2022-09-16
enlarge the space for seeds for simulations

### LINKED ISSUE #338: only few seeds possible in simulations
**Summary**
In https://github.com/Roche/crmPack/blob/main/R/Design-methods.R#L213 effectively *all* seeds for any simulations are restricted to a number 1:1e5.  

This limits simulation runs to 1e5 iterations which might not always suffice these days. Secondly, when a critical case gets studied in further detail, **all** simulations have to start from one of these seeds. Two simulations with same parameters and length 1e4 will nearly do 1000 identical simulation runs, biasing results.

reikoch:
# Pull Request

Fixes #338

### DIFF
--- a/R/Design-methods.R
+++ b/R/Design-methods.R
@@ -210,7 +210,7 @@ setMethod("simulate",
 
               ## from this,
               ## generate the individual seeds for the simulation runs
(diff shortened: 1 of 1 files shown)


danielinteractive (approved):
Thanks Reinhold!!

PR #355, opened 2022-09-27
Closes #352

### LINKED ISSUE #352: `StoppingMTDdistribution` class cleanup.

To do:

- [ ] Revise documentation
  - [ ] Replace `##'` by `#'`
  - [ ] Add to `pkgdown.yml`
- [ ] Add unit tests
  - [ ] For classes:
  - [ ] For functions: as usual
- [ ] Refactor as needed
  - [ ] Remove `keywords` tags
  - [ ] For classes:

wwojciech:
Closes #352


R/Rules-class.R
```diff
@@ -1331,71 +1331,64 @@ StoppingTargetProb <- function(target,
...
+## class ----
 
-                     o$result()
-                 })
-validObject(.StoppingMTDdistribution())
+#' `StoppingMTDdistribution`
+#'
+#' @description `r lifecycle::badge("stable")`
+#'
+#' [`StoppingMTDdistribution`] is the class for stopping based on MTD distribution.
```
danielinteractive, line 1342:
```suggestion
#' [`StoppingMTDdistribution`] is the class for stopping based on the posterior distribution of the MTD.
```

R/Rules-class.R
```diff
@@ -1331,71 +1331,64 @@ StoppingTargetProb <- function(target,
...
+#' It is used for the cases where the stopping occurs when there exists
+#' `prob` probability above a threshold of `tresh` of the current MTD been
+#' reached. Here, the MTD is defined as the dose for which `prob(DLT) = target`.
```
danielinteractive, line 1345:
```suggestion
#' reached. Here, the MTD is defined as the dose for which `prob(DLT)` is closest to the target.
```
because we don't get equality to any number?

danielinteractive: hm can you please check the logic used in https://github.com/Roche/crmPack/blob/main/R/Rules-methods.R and then reword this accordingly? I think the original wording was already misleading

danielinteractive (commented):
thanks Wojciech, please see inside

R/Rules-class.R
wwojciech: > because we don't get equality to any number?

Thanks @danielinteractive, hmm, I think this is the equality (excluding any differences resulting from finite arithmetic precisions issues). See the extract from the corresponding method function:
```
mtdSamples <- dose(x = stopping@target, model, samples)
 ```

Comment (danielinteractive) [reply R/Rules-class.R:1345]: but I guess `dose()` is doing this via closest one and not via equality right? Otherwise we would need to have infinitely fine dose grid

Comment (wwojciech) [reply R/Rules-class.R:1345]: > but I guess `dose()` is doing this via closest one and not via equality right? Otherwise we would need to have infinitely fine dose grid

I think the `dose()` computes the dose regardless the dose grid. See the example below.
```
setMethod(
  f = "dose",
  signature = signature(
    x = "numeric",
    model = "LogisticNormal",
    samples = "Samples"
  ),
  definition = function(x, model, samples) {
    assert_subset(c("alpha0", "alpha1"), names(samples@data))
    alpha0 <- samples@data$alpha0
    alpha1 <- samples@data$alpha1
    ref_dose <- as.numeric(model@ref_dose)
    assert_probabilities(x, len = h_null_if_scalar(alpha0))

    exp((logit(x) - alpha0) / alpha1) * ref_dose
  }
)
```

danielinteractive: hm ok maybe I have to read this again in detail. somehow the class description is not clear to me right now (also before)

wwojciech: Thanks @danielinteractive, I agree, I have just done some changes now, do you think it is ok now?

danielinteractive (approved):
ok looks good, thanks Wojciech!

PR #357, opened 2022-09-28
Closes #356

### LINKED ISSUE #356: `StoppingMTDCV` class cleanup.

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #356

### DIFF
--- a/R/Rules-validity.R
+++ b/R/Rules-validity.R
@@ -422,10 +422,6 @@ v_stopping_mtd_distribution <- function(object) {
   v$result()
 }
 
-#' @describeIn v_stopping validates that the [`StoppingLowestDoseHSRBeta`]
-#'  object contains valid probability target, threshold and shape parameters.
-v_stopping_lowest_dose_hsr_beta <- v_increments_hsr_beta
-
 #' @describeIn v_stopping validates that the [`StoppingMTDCV`] object
 #'   contains valid probability target and percentage threshold.
 v_stopping_mtd_cv <- function(object) {
@@ -435,8 +431,12 @@ v_stopping_mtd_cv <- function(object) {
     "target must be probability value from (0, 1) interval"
   )
   v$check(
-    test_probability(object@thresh_cv / 100, bounds_closed = FALSE),
+    test_probability(object@thresh_cv / 100, bounds_closed = c(FALSE, TRUE)),
     "thresh_cv must be percentage > 0"
   )
   v$result()
 }
+
+#' @describeIn v_stopping validates that the [`StoppingLowestDoseHSRBeta`]
+#'  object contains valid probability target, threshold and shape parameters.
+v_stopping_lowest_dose_hsr_beta <- v_increments_hsr_beta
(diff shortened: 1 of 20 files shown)


R/Rules-validity.R
```diff
@@ -435,8 +431,12 @@ v_stopping_mtd_cv <- function(object) {
     "target must be probability value from (0, 1) interval"
```
danielinteractive, line 431: need to adapt this?
```suggestion
    "target must be probability value from (0, 1] interval"
```

tests/testthat/test-Rules-class.R
```diff
@@ -422,18 +422,25 @@ test_that("StoppingMTDdistribution object can be created with user constructor",
...
-
-test_that("StoppingMTDCV object can be created with user constructor", {
+test_that("StoppingMTDCV object can be created with user constructor (default)", {
   result <- expect_silent(StoppingMTDCV())
   expect_valid(result, "StoppingMTDCV")
+  expect_identical(result@target, 0.3)
+  expect_identical(result@thresh_cv, 40)
+})
+
+test_that("StoppingMTDCV object can be created with user constructor (de", {
```
danielinteractive, line 439:
```suggestion
test_that("StoppingMTDCV object can be created with user constructor (default)", {
```
?

tests/testthat/test-Rules-validity.R
```diff
@@ -880,48 +872,47 @@ test_that("v_stopping_mtd_distribution returns message for non-valid prob", {
...
 
-  expect_equal(
-    v_stopping_mtd_cv(object),
-    c(
-      "target must be probability value from (0, 1) interval",
-      "thresh_cv must be percentage > 0"
-    )
-  )
+  # Changing `target` so that it does not represent allowed probability value.
+  object@target <- 1
```
danielinteractive, line 885: I thought that should be ok?

danielinteractive (commented):
thanks Wojciech!

tests/testthat/test-Rules-class.R
wwojciech: Oh, that is a typo. Tx!

tests/testthat/test-Rules-validity.R
wwojciech: See the comment for the first conversation, above.

R/Rules-validity.R
```diff
@@ -435,8 +431,12 @@ v_stopping_mtd_cv <- function(object) {
...
   )
   v$check(
-    test_probability(object@thresh_cv / 100, bounds_closed = FALSE),
+    test_probability(object@thresh_cv / 100, bounds_closed = c(FALSE, TRUE)),
```
danielinteractive, line 434: OK I guess I confused with this here - maybe good if you double check that the tests for this are correct?

wwojciech: I think this is ok, so `tresh_cv = 100` means that we stop for sure, regardless of the posterior dose samples, as:
```
do_stop <- (mtd_cv <= stopping@thresh_cv)
```
and `mtd_cv` takes on value from `[0, 1]`.

danielinteractive (approved):
thanks I just double checked the tests you have for the `thresh_cv` and looks good

PR #366, opened 2022-09-29
Closes #358

### LINKED ISSUE #358: `test-Rules-validity` - add few more specific test.
A few more tests should be added for the edge case. For instance:

- [x] if a slot of a class stores a probability value without `0`, `1` bounds, there should be a tests checking if the validation function reports a proper error for these edge case.
- [x] Also, if `0` or `1` is allowed, then it should be checked that the validation function passes when user tries to specify these edge values.
- [x] Also adjust the help for classes for same cases.

wwojciech:
Closes #358

### DIFF
--- a/R/Rules-class.R
+++ b/R/Rules-class.R
@@ -35,7 +35,8 @@ setClass(
 #'
 #' [`NextBestMTD`] is the class for next best dose based on MTD estimate.
 #'
-#' @slot target (`proportion`)\cr target toxicity probability.
+#' @slot target (`proportion`)\cr target toxicity probability, except \eqn{0} or
+#'   \eqn{1}.
 #' @slot derive (`function`)\cr a function which derives the final next best MTD
 #'   estimate, based on vector of posterior MTD samples. It must therefore accept
 #'   one and only one argument, which is a numeric vector, and return a number.
@@ -99,7 +100,7 @@ NextBestMTD <- function(target, derive) {
 #' @slot overdose (`numeric`)\cr the overdose toxicity interval (lower limit
 #'   excluded, upper limit included). It is used to filter probability samples.
 #' @slot max_overdose_prob (`proportion`)\cr maximum overdose posterior
-#'   probability that is allowed.
+#'   probability that is allowed, except \eqn{0} or \eqn{1}.
 #'
 #' @aliases NextBestNCRM
 #' @export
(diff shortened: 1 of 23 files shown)


R/Rules-class.R
```diff
@@ -35,7 +35,8 @@ setClass(
 #'
 #' [`NextBestMTD`] is the class for next best dose based on MTD estimate.
 #'
-#' @slot target (`proportion`)\cr target toxicity probability.
+#' @slot target (`proportion`)\cr target toxicity probability, except \eqn{0} or
+#'   \eqn{1}.
```
danielinteractive, line 39: for maintenance purpose, highly recommend to just say 1 or 0 without the markup around it.

R/Rules-class.R
```diff
@@ -709,9 +712,9 @@ setClass(
 #' @slot intervals (`numeric`)\cr a vector with the left bounds of the relevant
 #'   intervals. This parameters specifies the right bounds of the intervals.
 #'   For example, `intervals  = c(0, 50, 100)` specifies three intervals:
-#'   (0, 50), [50, 100) and [100, +Inf). That means, the right bound of the
-#'   intervals are exclusive to the interval and the last interval goes from the
-#'   last value to infinity.
+#'   \eqn{(0, 50)}, \eqn{[50, 100)} and \eqn{[100, +Inf)}. That means, the right
```
danielinteractive, line 715: here it is useful and ok

R/Rules-class.R
```diff
@@ -761,21 +764,24 @@ IncrementsRelative <- function(intervals, increments) {
...
+#' dose in case of a DLT in part \eqn{1}, or `clean_start` shift (e.g. \eqn{0})
+#' in case of no DLTs in part \eqn{1}.
+#' If part \eqn{1} will still be on in the
+#' next cohort, then the next dose level will be the next higher dose level in
+#' the `part1Ladder` slot of the data object.
+#' If part \eqn{2} has been started before, the usual relative increment rules
+#' apply, see [`IncrementsRelative`].
+#'
+#' @slot dlt_start (`count`)\cr the dose level increment for starting part
+#'   \eqn{2} in case of a DLT in part \eqn{1}.
```
danielinteractive, line 779: also here, just 2 and 1

danielinteractive (commented):
thanks Wojciech, please see inside

R/Rules-class.R
wwojciech: ohhh, sure, then It would be a lot o changes now :)

R/Rules-class.R
wwojciech: acha ok.

R/Rules-class.R
danielinteractive: you can try find and replace across files, tip:

Replace a string with another one in multiple files in this directory:
`perl -pi -w -e 's/OldText/NewText/g;' *.txt`

danielinteractive (approved):

PR #368, opened 2022-09-30
Closes #367 Change rule from 1 to 2 to return always a value

### LINKED ISSUE #367: OneParExpNormalPrior - skel_fun_inv produce NA
The following code is also part of the JSS publication (Bove et al. 2019) which provides code for the one-parameter CRM (O’Quigley et al. 1990):

https://github.com/Roche/crmPack/blob/d7cb6def0d43ba7cffbb439b367feda027934991/R/Model-class.R#L2621-L2622

Two interpolation functions are introduced to be able to get any DLT prob over the dose range and any dose for a given DLT probability. 

In the function skel_fun_inv the parameter rule=1 is specified. That means that NAs are returned when the prob is outside the range of skel_probs. The Problem is that crmPack cannot deal with NAs in some context (e.g. next best dose) and therefore simulations runs into errors.

I think that rule=2 fix the problem (lowest or highest dose is returned when prob outside skel_probs), return what we want and is in line with the behavior of crmPack (to always return a dose).

I am not sure if there was a reason to choose rule=1 in the JSS paper, therefore I also ask @TJaki for input to this issue.

0liver0815:
Closes #367

### DIFF
--- a/R/Model-class.R
+++ b/R/Model-class.R
@@ -2619,7 +2619,7 @@ OneParExpNormalPrior <- function(skel_probs,
   assert_number(sigma2, lower = .Machine$double.xmin, finite = TRUE)
 
   skel_fun <- approxfun(x = dose_grid, y = skel_probs, rule = 2)
-  skel_fun_inv <- approxfun(x = skel_probs, y = dose_grid, rule = 1)
+  skel_fun_inv <- approxfun(x = skel_probs, y = dose_grid, rule = 2)
 
   .OneParExpNormalPrior(
     skel_fun = skel_fun,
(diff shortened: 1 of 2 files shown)


danielinteractive (approved):

danielinteractive:
@0liver0815 thanks again, if possible it would be best to even have a small test that fails because there is NA in the current version, and then shows that it works now

danielinteractive:
@0liver0815 I think here just a test is missing?

0liver0815:
@danielinteractive I am working on a test. I think I will commit today. Then the request can be merged.

0liver0815:
@danielinteractive Test case added.

PR #369, opened 2022-09-30
Closes #363 assert_true checks not correct. After discussion the whole method for…

### LINKED ISSUE #363: mcmc-GeneralData-OneParExpNormalPrior - assert check
There seems to be an error in the mcmc method and the correspnding description: 

https://github.com/Roche/crmPack/blob/d7cb6def0d43ba7cffbb439b367feda027934991/R/mcmc.R#L237-L239

https://github.com/Roche/crmPack/blob/d7cb6def0d43ba7cffbb439b367feda027934991/R/mcmc.R#L259-L260

The description should read:
```R
#'   [`OneParExpNormalPrior`] model, it is required that the length of
#'   skeleton prior probabilities vector should be equal to the length of the number of doses
```
and the code
```R
  if (!from_prior) {
      assert_true(length(model@skel_probs) == max(data@nGrid))
```
I am wondering if this check should not be done when the model is specified with the other model parameter validation
checks.

0liver0815:
closes #363 

… OneParExpNormalPrior can be removed as not needed.

Tested with simulation. Works without the method in mcmc.

### DIFF
--- a/R/mcmc.R
+++ b/R/mcmc.R
@@ -257,7 +257,7 @@ setMethod(
                  from_prior = data@nObs == 0L,
                  ...) {
     if (!from_prior) {
-      assert_true(length(model@skel_probs) == max(data@nGrid))
+      assert_true(length(model@skel_probs) == data@nGrid)
     }
 
     callNextMethod(
--- a/R/mcmc.R
+++ b/R/mcmc.R
@@ -235,8 +235,8 @@ setMethod(
 
 #' @describeIn mcmc Standard method which uses JAGS. For the
 #'   [`OneParExpNormalPrior`] model, it is required that the length of
-#'   skeleton prior probabilities vector should be equal to the maximum dose
-#'   level observed in data.
+#'   skeleton prior probabilities vector should be equal to the length of the
+#'   number of doses.
 #'
 #' @param from_prior (`flag`)\cr sample from the prior only? Default to `TRUE`
 #'   when number of observations in `data` is `0`. For some models it might be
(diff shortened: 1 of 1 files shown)


danielinteractive (commented):
Thanks @0liver0815 , can you add a test as well where this is simulated, i.e. standard `mcmc()` is run for this model? or is it already included in codebase?

0liver0815:
@danielinteractive Yes it runs, but after thinking about the check of the slots in the model, I now believe that this check here is necessary. The reason is that it is possible that the user specifies the grid in the model, but the data may not fit to check and mcmc crash. So blease don't merge, I will change the code back and fix the check.

danielinteractive:
@0liver0815 right, makes sense now. basically dose_grid is defined in the data outside the model therefore no guarantee that they match together well via the skeleton...

danielinteractive:
@0liver0815 what is the next step here?

0liver0815:
@danielinteractive From my point of view this branch can already be merged. I can also add a test, before merging or the test can be added later.

R/mcmc.R
```diff
@@ -257,7 +257,7 @@ setMethod(
                  from_prior = data@nObs == 0L,
                  ...) {
     if (!from_prior) {
-      assert_true(length(model@skel_probs) == max(data@xLevel))
+      assert_true(length(model@skel_probs) == max(data@nGrid))
```
danielinteractive, line 260:
```suggestion
      assert_true(length(model@skel_probs) == data@nGrid)
```

danielinteractive (approved):
just one simplification inside, afterwards good to merge

PR #371, opened 2022-10-02
Closes #361

### LINKED ISSUE #361: OneParExpNormalPrior validation function runs into error for valid parameter combinations
With the code
```
my_model <- OneParExpNormalPrior(
  skel_probs = seq(from = 0.1, to = 0.7, length = 5),
  dose_grid = 1:5,
  sigma2 = 2
)
```
the following error message was produced
```
Error in validObject(.Object) : 
  invalid class “OneParExpNormalPrior” object: skel_fun_inv must be an inverse funtion of skel_fun function 
```

This error is generate by the validation
https://github.com/Roche/crmPack/blob/d7cb6def0d43ba7cffbb439b367feda027934991/R/Model-validity.R#L516

After discussion with @wwojciech this Issue is added as it needs further investigation of the root cause.

wwojciech:
Closes #361

### DIFF
--- a/.lintr
+++ b/.lintr
@@ -1,4 +1,4 @@
-linters: with_defaults(
+linters: linters_with_defaults(
   line_length_linter = line_length_linter(120),
   cyclocomp_linter = NULL,
   object_usage_linter = NULL
--- a/R/Model-validity.R
+++ b/R/Model-validity.R
@@ -510,11 +510,11 @@ v_model_tite_logistic_log_normal <- function(object) {
 #' @describeIn v_model_objects validates that [`OneParExpNormalPrior`] class slots are valid.
 v_model_one_par_exp_normal_prior <- function(object) {
   v <- Validate()
-  y <- seq(from = 0, to = 1, by = 0.1) # Skeleton prior probabilities.
+  y <- seq(from = 0, to = 1, by = 0.1) # Probabilities.
   x <- object@skel_fun_inv(y) # Dose grid.
   not_NA <- !is.na(x)
   v$check(
-    all(object@skel_fun(x[not_NA]) == y[not_NA]),
+    isTRUE(all.equal(object@skel_fun(x[not_NA]), y[not_NA])),
     "skel_fun_inv must be an inverse funtion of skel_fun function"
   )
   v$check(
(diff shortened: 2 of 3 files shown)


wwojciech:
Hi @cicdguy , Do you know what happened that `Linter` complains now about function names longer than 30 characters? It did not complain before....
```
Function with_defaults was deprecated in lintr version 3.0.0. Use linters_with_defaults or modify_defaults instead. 
Warning: file=/github/workspace/R/Model-validity.R,line=49,col=1,[object_length_linter] Variable and function names should not be longer than 30 characters.
```

wwojciech:
> > Hi @cicdguy , Do you know what happened that `Linter` complains now about function names longer than 30 characters? It did not complain before....
> > ```
> > Function with_defaults was deprecated in lintr version 3.0.0. Use linters_with_defaults or modify_defaults instead. 
> > Warning: file=/github/workspace/R/Model-validity.R,line=49,col=1,[object_length_linter] Variable and function names should not be longer than 30 characters.
> > ```
> 
> Lints are applied to files that have been modified. Even if you haven't modified the code, as long as the _file_ has been modified, linter errors are determined for the file. You should be able to assess all lints across the codebase using the `lintr` package. I see `4017` style-related lint errors in this entire codebase, which means that you will see issues more often than so.

Thank you @cicdguy Yap, this one I know, as it happens quite often when we modify some old file. But actually, this is a new file we created some time ago (when the whole `crmPack` was already under `CI/CD` "supervision", and it was checked every time it was updated with `linter`, also on my local computer.

wwojciech:
Hi @danielinteractive I think that this `lintr` behaviour is very strange, as `Model-validity.R` is a new file we created as we started to clean `crmPack` and it was checked against `lintr` every time we modify it. I did not know why it started to complain about some old code now, and not before. However, I see we can either:
1. disable linter check for the non-complaint lines, or
2. change functions names (this would be a bit of time consuming).
What do you think?
Thanks a lot!

danielinteractive (approved):

wwojciech:
> Hi @danielinteractive I think that this `linter` behaviour is very strange, as `Model-validity.R` is a new file we created as we started to clean `crmPack` and it was checked against `linter` every time we modify it. I did not know why it started to complain about some old code now, and not before. However, I see we can either:
> 
> 1. disable linter check for the non-complaint lines, or
> 2. change functions names (this would be a bit of time consuming), or
> 3. we add an exception to `linter` config file so that it accepts a bit longer function names (as this is the main source of `linter` violations.
> 
> What do you think? Thanks a lot!

Hi @danielinteractive - would you have some comments on this? Tx!

wwojciech:
I will merge this branch for now, and create a separate issue for the `lintr` problems.

PR #376, opened 2022-10-04
Closes #375

### LINKED ISSUE #375: New `lintr` version 3.0.0 seems to be more strict.
Seems that the new version of `lintr` is now very strict and it started to triggers flags for the code for which it did not throw it before. See e.g. https://github.com/Roche/crmPack/pull/371 , where the problem appears for Model-validity.R` which is a new file that we created as we started to clean `crmPack` and it was checked against `linter` every time we modify it. It now complains about the some old code, about which it did not complain before....We can either:
 
 1. disable linter check for the non-complaint lines, or
 2. change functions names (this would be a bit of time consuming), or
 3. we add an exception to `linter` config file so that it accepts a bit longer function names (as this is the main source of `linter` violations.

wwojciech:
Closes #375


NAMESPACE
```diff
@@ -3,6 +3,102 @@
 S3method(plot,gtable)
 S3method(print,gtable)
 export("%~%")
+export(.CohortSizeConst)
```
danielinteractive, line 6: why are the exports of default constructors needed for linter fix?

R/Model-class.R
```diff
@@ -1145,7 +1145,7 @@ DualEndpoint <- function(mean,
   assert_numeric(rho, min.len = 1, max.len = 2)
 
   use_fixed <- c(sigma2W = is.scalar(sigma2W), rho = is.scalar(rho))
-  betaZ_params <- ModelParamsNormal(mean, cov)
+  betaZ_params <- ModelParamsNormal(mean, cov) # nolintr
```
danielinteractive, line 1148: ok because I guess we have `betaZ` as parameter so we don't want to change that

wwojciech:
Hi @danielinteractive - just a note here: I left some internal variable names unchanged because they somehow refer to the variables in mathematical equations.

R/Model-class.R
```diff
@@ -2284,11 +2284,11 @@ EffFlexi <- function(eff,
   D1 <- cbind(0, diag(data@nGrid - 1)) - cbind(diag(data@nGrid - 1), 0)
   if (rw1) { # the rank-deficient prior precision for the RW1 prior.
     RW <- crossprod(D1)
-    RW_rank <- data@nGrid - 1L # rank = dimension - 1.
+    RW_rank <- data@nGrid - 1L # rank = dimension - 1. # nolintr
```
danielinteractive, line 2287: I think here it would be nice to update to `rw` etc. since that does not need any changes elsewhere?

R/Model-validity.R
```diff
@@ -79,7 +79,7 @@ v_model_logistic_normal_mix <- function(object) {
 
 #' @describeIn v_model_objects validates that `component` is a list with
 #'   valid `ModelParamsNormal` objects as well as `weights` are correct.
-v_model_logistic_normal_fixed_mix <- function(object) {
+v_model_logistic_normal_fixed_mix <- function(object) { # nolintr
```
danielinteractive, line 82: is there a threshold we can increase for the accepted name length? I thought so?

R/Model-validity.R
```diff
@@ -508,13 +508,13 @@ v_model_tite_logistic_log_normal <- function(object) {
 }
 
 #' @describeIn v_model_objects validates that [`OneParExpNormalPrior`] class slots are valid.
-v_model_one_par_exp_normal_prior <- function(object) {
+v_model_one_par_exp_normal_prior <- function(object) { # nolintr
   v <- Validate()
   y <- seq(from = 0, to = 1, by = 0.1) # Probabilities.
   x <- object@skel_fun_inv(y) # Dose grid.
-  not_NA <- !is.na(x)
+  notNA <- !is.na(x)
```
danielinteractive, line 515:
```suggestion
  not_na <- !is.na(x)
```
would also be good

R/Rules-methods.R
```diff
@@ -383,8 +383,8 @@ setMethod(
 
     # Get number of patients per grid's dose and DLT rate at the last level.
     nPatients <- table(factor(data@x, levels = data@doseGrid))
-    nDLTs_last_level <- sum(data@y[data@xLevel == last_level])
-    DLT_rate_last_level <- nDLTs_last_level / nPatients[last_level]
+    nDLTs_last_level <- sum(data@y[data@xLevel == last_level]) # nolintr
```
danielinteractive, line 386:
```suggestion
    n_dlts_last_level <- sum(data@y[data@xLevel == last_level])
```
?

danielinteractive (commented):
thanks Wojciech, please see inside, generally I would try to conform with the patterns, only in the case where we need to many updates outside of the current function scope e.g. then stick to non-standard patterns

R/Model-class.R
wwojciech: Exactly.

R/Model-class.R
wwojciech: yah, just wanted to keep upper case for matrices, since RW is a matrix, also mentioned in the help, otherwise it might be mixed with `rw1` flag we use also here.....

R/Model-validity.R
wwojciech: Yes, there is, it is currently set to 30, and if I set 35 then it covers these cases, but I though that 30 is a good limit and actually we should avoid longer functions, except these few exceptions.

R/Model-class.R
danielinteractive: ok up to you.. also `rw_mat` could be an option. basically the more we can minimize `# nolintr` usage the easier for maintenance as well

NAMESPACE
wwojciech: This is not related, I just had to run the `Document` `roxygen` so that the help pages are updated after `lintr` changes and then the `NAMESPACE` file was changed (as it is maintained by the `roxygen`).

R/Model-validity.R
danielinteractive: hm ok as well... if it is really only less than 5 exceptions in whole package... otherwise either change names or increase threshold

R/Model-class.R
wwojciech: Thanks @danielinteractive , I agree. Maybe in this case I would leave `RW` as it corresponds with the name of the slot in the corresponding class, so I would need to change the class itself. Maybe it would be also a good idea but we then have other variables-matrices, where we do not use `_mat` suffix, and these come from the equation and etc. So I would just leave it here as it is.

NAMESPACE
wwojciech: I do not know why it was not exported before, even that the `@export` tag was there before. We can maybe remove the `@export` tag from the default constructors.

danielinteractive (approved):

PR #394, opened 2022-10-13
Correct validity check

### LINKED ISSUE #393: Error in v_model_objects validates that [`OneParExpNormalPrior`] class slots are valid
After error correction #367 it turns out that the validity checks for the slots do not work as specified

https://github.com/Roche/crmPack/blob/8d913a71ebfab84df35e6fdd7aad0bbf694d8f0b/R/Model-validity.R#L511-L519

As far as I can see, the problem is that fixed sequence of 11 probs is generated to check if the inverse function is correct. The problem after merging #367 is that no NA are returned anymore, but the highest or lowest dose in case that the prob is outside of the range of the skeleton probs.
I think the check will now fail anytime when the fixed sequence does not match the skeleton probs. I think the following code will solve the problem:
```
# user specified values:
dose_grid <- 1:11
skel_probs <- seq(0.1, 0.5, length.out=11)

# just helper to test easily (not necessary in model-validity.r
skel_fun <- approxfun(x = dose_grid, y = skel_probs, rule = 2)
skel_fun_inv <- approxfun(x = skel_probs, y = dose_grid, rule = 2)

# new test code:
x <- skel_fun_inv(skel_probs) # Dose grid.
all.equal(skel_fun(x), skel_probs)
```

0liver0815:
Closes #393

### DIFF
--- a/R/Model-validity.R
+++ b/R/Model-validity.R
@@ -510,11 +510,9 @@ v_model_tite_logistic_log_normal <- function(object) { # nolintr
 #' @describeIn v_model_objects validates that [`OneParExpNormalPrior`] class slots are valid.
 v_model_one_par_exp_normal_prior <- function(object) { # nolintr
   v <- Validate()
-  y <- seq(from = 0, to = 1, by = 0.1) # Probabilities.
-  x <- object@skel_fun_inv(y) # Dose grid.
-  not_na <- !is.na(x)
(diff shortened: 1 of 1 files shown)


wwojciech:
Hi @0liver0815!
Lets solve this issue via https://github.com/Roche/crmPack/pull/396 PR. Please do not merge as there are actually much more changes required. I will close this PR to avoid any confusions.

PR #396, opened 2022-10-13
Closes #393

### LINKED ISSUE #393: Error in v_model_objects validates that [`OneParExpNormalPrior`] class slots are valid

wwojciech:
Closes #393

### DIFF
--- a/tests/testthat/test-Model-validity.R
+++ b/tests/testthat/test-Model-validity.R
@@ -1060,12 +1060,15 @@ test_that("v_model_one_par_exp_normal_prior passes for valid object (finit art. 
 test_that("v_model_one_par_exp_normal_prior returns message for wrong skel_fun - skel_fun_inv", {
   object <- h_get_one_par_exp_normal_prior()
 
-  # Assigning wrong skel_probs.
+  # Assigning wrong skel_fun/skel_fun_inv
   object@skel_fun <- function(x) 2 * x
   object@skel_fun_inv <- function(x) x^2
   expect_equal(
     v_model_one_par_exp_normal_prior(object),
-    "skel_fun_inv must be an inverse funtion of skel_fun function"
+    c(
+      "skel_fun_inv must be an inverse funtion of skel_fun function on within the range of sekeleton probs",
+      "skel_fun_inv must be an inverse funtion of skel_fun function on outside the range of sekeleton probs"
+    )
   )
 })
 
@@ -1076,7 +1079,14 @@ test_that("v_model_one_par_exp_normal_prior returns message for wrong skel_probs
   object@skel_probs <- c(-1, 0.5, Inf)
   expect_equal(
     v_model_one_par_exp_normal_prior(object),
-    "skel_probs must be probabilities between 0 and 1"
+    "skel_probs must be sorted probability values between 0 and 1"
+  )
+
+  # Assigning not sorted skel_probs.
+  object@skel_probs <- c(0.2, 0.1)
+  expect_equal(
+    v_model_one_par_exp_normal_prior(object),
+    "skel_probs must be sorted probability values between 0 and 1"
   )
 })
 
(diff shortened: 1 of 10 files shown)


danielinteractive (commented):
looks good to me, would be good if @0liver0815 also tries it out interactively maybe

wwojciech:
Hi @danielinteractive , @0liver0815 It is ready to review now. I cannot make assertion on `skel_probs` for uniqueness, as we do not have this ready in `assert_probabilities`. So I will first need to enhance `assert_probabilities` first and then add it here (this will be done within a separated issue https://github.com/Roche/crmPack/issues/397). But I think it is ok for now to proceed without it.

tests/testthat/test-Model-validity.R
```diff
@@ -1079,14 +1079,14 @@ test_that("v_model_one_par_exp_normal_prior returns message for wrong skel_probs
   object@skel_probs <- c(-1, 0.5, Inf)
   expect_equal(
     v_model_one_par_exp_normal_prior(object),
-    "skel_probs must be sorted probability values between 0 and 1"
+    "skel_probs must be a unique sorted probability values between 0 and 1"
   )
 
   # Assigning not sorted skel_probs.
   object@skel_probs <- c(0.2, 0.1)
```
danielinteractive, line 1086: what happens with `c(0.2, 0.2)`? is that caught?

wwojciech: Unfortunately not yet, as I have to enhance `assert_probabilities` to be able to check for the uniqueness. I just prepared these error messages already.

wwojciech: but maybe we can actually allow repeated values. The function will not be strictly incr. then, but since we have this interpolation - then it can be fine still. What do you think @danielinteractive ? Dose it make sense practically, that we have the same DLT prob for different doses?

danielinteractive: yeah but then it is not a bijective function?

wwojciech: Yeah, that is right, sorry I do not why I though that this wont be a problem with this interpolation. But this is obviously  wrong. So we have to add it.

wwojciech: Would it be ok if you accept it so that these unit tests errors disappear from the `main` and then I will work on the `assert_probabilities`? Tx!

danielinteractive: ok, just make a new issue for this then

danielinteractive (approved):

tests/testthat/test-Model-validity.R
wwojciech: Sure, thanks. Created - https://github.com/Roche/crmPack/issues/399

Issue #323, opened 2022-09-09
Design broom-like `tidy()` and `glimpse()` methods


danielinteractive:
To do:

- [x] Add design doc in `design` folder
- [x] Describe idea
- [ ] Think about potentially different solution paths
- [ ] Discuss questions such as 
	- [ ] whether to include attributes of the whole object in each row of the data frame
- [ ] Add prototype code

PuzzledFace:
Some classes, such as `StioppingRuleTargetProb`, specify ranges in unnamed vectors:

```
StoppingTargetProb(target=c(0.2, 0.35), prob=0.5)
```
Currently, the `tidy()` generic converts this to:
```
# A tibble: 2 × 3
  Class              target  prob
  <chr>               <dbl> <dbl>
1 StoppingTargetProb   0.2    0.5
2 StoppingTargetProb   0.35   0.5
```

Would this be better represented as

```
# A tibble: 1 × 4
  Class               prob   min   max
  <chr>              <dbl> <dbl> <dbl>
1 StoppingTargetProb   0.5   0.2  0.35
```

I think yes.

Here's a draft list of classes that are potentially affected:

```
StoppingtargetBiomarker
StoppingTargetProb
```

danielinteractive:
Is now being worked on in #587 therefore closing here

PR #408, opened 2022-10-21
323 design broom like tidy and glimpse methods

PuzzledFace:
Implementing `tidy()` methods for `crmPack` classes.  First part of #323.

Please review and comment.


danielinteractive:
thanks @PuzzledFace - for this stage, design, can you try to only add a design doc and not yet change the package itself? that will make the review much easier

danielinteractive:
@PuzzledFace could you try to only include the design doc on this one?

danielinteractive:
Hi @PuzzledFace , it would be important to first have the design step here, meaning just a single Rmd document in design folder, otherwise I will not be able to review this - is that possible?

danielinteractive:
Hi @PuzzledFace could we just have the single design doc first? Otherwise I cannot see the scope. (Generally, design PRs should only change very few files and in most cases only modify or add a design doc.)

PR #431, opened 2022-11-17
Closes #430

### LINKED ISSUE #430: `CohortSizeParts` rules class cleanup

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #430


R/Rules-class.R
```diff
@@ -2057,53 +2057,46 @@ CohortSizeConst <- function(size) {
...
-                             "2 elements required in sizes")
+#' `CohortSizeParts`
+#'
+#' @description `r lifecycle::badge("stable")`
+#'
+#' [`CohortSizeParts`] is the class for cohort size that changes for the second
+#' part of the dose escalation. It works only in conjunction with [`DataParts`]
+#' objects.
+#'
+#' @slot sizes (`integer`)\cr a vector of length two with two sizes, one for
```
danielinteractive, line 2072:
```suggestion
#' @slot sizes (`numeric`)\cr a vector of length two with two sizes, one for
```
?

danielinteractive (approved):

R/Rules-class.R
wwojciech: Hi @danielinteractive - why `numeric`? According to NEST data types, `integer` is ok.

```
numeric | numeric vector (integer or double)
number | numeric scalar (integer or double)
proportion | number between 0 and 1 (exclusive)
integer | integer vector
```

danielinteractive: alright

PR #438, opened 2022-11-21
321 add o quigley model from jss paper

### LINKED ISSUE #321: Add “O’Quigley arctan model” from the JSS paper into the package
To do:

- [ ] Take classes and methods from the JSS paper into the package
- [ ] add tests

JSS Paper: https://www.jstatsoft.org/article/view/v089i10

0liver0815:
Add one parameter exponential model with exponential prior from JSS paper. Closes #321.

### DIFF
--- a/R/Model-class.R
+++ b/R/Model-class.R
@@ -2662,13 +2662,14 @@ OneParExpNormalPrior <- function(skel_probs,
 #' @description `r lifecycle::badge("experimental")`
 #'
 #' [`OneParExpExpPrior`] is the class for a standard CRM with an exponential prior on
-#' the log power parameter for the skeleton prior probabilities.
+#' the log power parameter for the skeleton prior probabilities. It is an
+#' implementation of a version of the one-parameter CRM (O’Quigley et al. 1990).
 #'
 #' @slot skel_fun (`function`)\cr function to calculate the prior DLT probabilities.
 #' @slot skel_fun_inv (`function`)\cr inverse function of `skel_fun`.
 #' @slot skel_probs (`numeric`)\cr skeleton prior probabilities. This is a vector
 #'   of unique and sorted probability values between 0 and 1.
-#' @slot lambda (`number`)\cr prior rate parameter of of log power parameter alpha.
+#' @slot lambda (`number`)\cr rate parameter of prior exponential distribution for theta.
 #'
 #' @aliases OneParExpExpPrior
 #' @export
@@ -2693,14 +2694,14 @@ OneParExpNormalPrior <- function(skel_probs,
 #'   of unique and sorted probability values between 0 and 1.
 #' @param dose_grid (`numeric`)\cr dose grid. It must be must be a sorted vector
 #'   of the same length as `skel_probs`.
-#' @param lambda (`number`)\cr prior rate parameter of of log power parameter alpha.
+#' @param lambda (`number`)\cr rate parameter of prior exponential distribution for theta.
 #'
 #' @export
 #' @example examples/Model-class-OneParExpExpPrior.R
 #'
 OneParExpExpPrior <- function(skel_probs,
-                                 dose_grid,
-                                 lambda) {
+                              dose_grid,
+                              lambda) {
   assert_probabilities(skel_probs, unique = TRUE, sorted = TRUE) # So that skel_fun_inv exists.
   assert_numeric(dose_grid, len = length(skel_probs), any.missing = FALSE, unique = TRUE, sorted = TRUE)
   assert_number(lambda, lower = .Machine$double.xmin, finite = TRUE)
@@ -2715,12 +2716,12 @@ OneParExpExpPrior <- function(skel_probs,
     lambda = lambda,
     datamodel = function() {
       for (i in 1:nObs) {
-        p[i] <- skel_probs[xLevel[i]]^exp(alpha)
+        p[i] <- skel_probs[xLevel[i]]^theta
         y[i] ~ dbern(p[i])
       }
     },
     priormodel = function() {
-      alpha ~ dexp(lambda)
+      theta ~ dexp(lambda)
     },
     modelspecs = function(from_prior) {
       ms <- list(lambda = lambda)
@@ -2730,10 +2731,10 @@ OneParExpExpPrior <- function(skel_probs,
       ms
     },
     init = function() {
-      list(alpha = 1)
+      list(theta = 1)
     },
     datanames = c("nObs", "y", "xLevel"),
-    sample = "alpha"
+    sample = "theta"
   )
 }
 
--- a/R/Model-class.R
+++ b/R/Model-class.R
@@ -2653,6 +2653,91 @@ OneParExpNormalPrior <- function(skel_probs,
   )
 }
 
+# OneParExpExpPrior ----
+
+## class ----
+
+#' `OneParExpExpPrior`
+#'
+#' @description `r lifecycle::badge("experimental")`
+#'
+#' [`OneParExpExpPrior`] is the class for a standard CRM with an exponential prior on
+#' the log power parameter for the skeleton prior probabilities.
+#'
+#' @slot skel_fun (`function`)\cr function to calculate the prior DLT probabilities.
+#' @slot skel_fun_inv (`function`)\cr inverse function of `skel_fun`.
+#' @slot skel_probs (`numeric`)\cr skeleton prior probabilities. This is a vector
+#'   of unique and sorted probability values between 0 and 1.
+#' @slot lambda (`number`)\cr prior rate parameter of of log power parameter alpha.
+#'
+#' @aliases OneParExpExpPrior
+#' @export
+#'
+.OneParExpExpPrior <- setClass(
+  Class = "OneParExpExpPrior",
+  slots = c(
+    skel_fun = "function",
+    skel_fun_inv = "function",
+    skel_probs = "numeric",
+    lambda = "numeric"
+  ),
+  contains = "GeneralModel",
+  validity = v_model_one_par_exp_exp_prior
+)
+
+## constructor ----
+
+#' @rdname OneParExpExpPrior-class
+#'
+#' @param skel_probs (`numeric`)\cr skeleton prior probabilities. This is a vector
+#'   of unique and sorted probability values between 0 and 1.
+#' @param dose_grid (`numeric`)\cr dose grid. It must be must be a sorted vector
+#'   of the same length as `skel_probs`.
+#' @param lambda (`number`)\cr prior rate parameter of of log power parameter alpha.
+#'
+#' @export
+#' @example examples/Model-class-OneParExpExpPrior.R
+#'
+OneParExpExpPrior <- function(skel_probs,
+                                 dose_grid,
+                                 lambda) {
+  assert_probabilities(skel_probs, unique = TRUE, sorted = TRUE) # So that skel_fun_inv exists.
+  assert_numeric(dose_grid, len = length(skel_probs), any.missing = FALSE, unique = TRUE, sorted = TRUE)
+  assert_number(lambda, lower = .Machine$double.xmin, finite = TRUE)
+
+  skel_fun <- approxfun(x = dose_grid, y = skel_probs, rule = 2)
+  skel_fun_inv <- approxfun(x = skel_probs, y = dose_grid, rule = 2)
+
+  .OneParExpExpPrior(
+    skel_fun = skel_fun,
+    skel_fun_inv = skel_fun_inv,
+    skel_probs = skel_probs,
+    lambda = lambda,
+    datamodel = function() {
+      for (i in 1:nObs) {
+        p[i] <- skel_probs[xLevel[i]]^exp(alpha)
+        y[i] ~ dbern(p[i])
+      }
+    },
+    priormodel = function() {
+      alpha ~ dexp(lambda)
+    },
+    modelspecs = function(from_prior) {
+      ms <- list(lambda = lambda)
+      if (!from_prior) {
+        ms$skel_probs <- skel_probs
+      }
+      ms
+    },
+    init = function() {
+      list(alpha = 1)
+    },
+    datanames = c("nObs", "y", "xLevel"),
+    sample = "alpha"
+  )
+}
+
+
 # FractionalCRM ----
 
 ## class ----
--- a/R/Model-methods.R
+++ b/R/Model-methods.R
@@ -714,12 +714,12 @@ setMethod(
     samples = "Samples"
   ),
   definition = function(x, model, samples) {
-    assert_subset("alpha", names(samples@data))
-    alpha <- samples@data$alpha
+    assert_subset("theta", names(samples@data))
+    theta <- samples@data$theta
     skel_fun_inv <- model@skel_fun_inv
-    assert_probabilities(x, len = h_null_if_scalar(alpha))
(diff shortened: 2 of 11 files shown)


danielinteractive:
@wwojciech maybe can you review as a first pass? Note that the general approach was agreed already (separate class with this name, etc.)

wwojciech:
> Collaborator

Sure @danielinteractive, with a pleasure.

wwojciech:
Hi @0liver0815 , I thinks validation function and mcmc are missing from the pckg-docs.

0liver0815:
I think the validation function `v_model_one_par_exp_exp_prior` is already there. I added `mcmc-GeneralData-OneParExpExpPrior`. However, I missed that function, because `mcmc-GeneralData-OneParExpNormalPrior` was also not included in `_pkdown.yaml`. I am wondering why only  mcmc-`GeneralData-OneParExpExpPrior` may causes an error as the whole code is very similar.

Finally adding `mcmc-GeneralData-OneParExpExpPrior` is not solving the error. I am a liitle bit lost, what I need to add.

0liver0815:
@wwojciech Finally I just did not run the documentation, which causes the errors. I think its now resolved.

R/Model-methods.R
```diff
@@ -698,6 +698,31 @@ setMethod(
   }
 )
 
+## OneParExpExpPrior ----
+
+#' @describeIn dose compute the dose level reaching a specific target
+#'   probability of the occurrence of a DLE (`x`).
```
wwojciech, line 704: Lets use `DLT` instead of `DLE` and remove that (`x`). Please also change it for the twin class.

R/Model-class.R
```diff
@@ -2653,6 +2653,92 @@ OneParExpNormalPrior <- function(skel_probs,
...
+#' @export
+#' @example examples/Model-class-OneParExpExpPrior.R
+#'
+OneParExpExpPrior <- function(skel_probs,
+                              dose_grid,
+                              lambda) {
+  assert_probabilities(skel_probs, unique = TRUE, sorted = TRUE) # So that skel_fun_inv exists.
+  assert_numeric(dose_grid, len = length(skel_probs), any.missing = FALSE, unique = TRUE, sorted = TRUE)
+  assert_number(lambda, lower = .Machine$double.xmin, finite = TRUE)
+
```
wwojciech, line 2708: Assertion for `lambda` is actually not needed here (we have it in class validation function). Same for `OneParExpNormalPrior` class and `sigma2` param.

R/Model-methods.R
```diff
@@ -698,6 +698,31 @@ setMethod(
...
+#'
+setMethod(
+  f = "dose",
+  signature = signature(
+    x = "numeric",
+    model = "OneParExpExpPrior",
+    samples = "Samples"
+  ),
+  definition = function(x, model, samples) {
+    assert_subset("theta", names(samples@data))
```
wwojciech, line 717: Would be good to also assert that `theta` is not zero (this may still theoretically happen as the support for the exp. dist. includes 0).

tests/testthat/test-Model-class.R
```diff
@@ -1094,6 +1094,120 @@ test_that("No NA is returned in dose calculations for OneParExpNormalPrior model
...
+test_that("OneParExpExpPrior object can be created with user constructor", {
+  result <- expect_silent(
+    OneParExpExpPrior(
+      skel_probs = seq(from = 0.1, to = 0.9, length = 5),
+      dose_grid = 1:5,
+      lambda = 2
+    )
+  )
+  expect_valid(result, "OneParExpExpPrior")
+})
```
wwojciech, line 1110: We should also test if the custom values for class slots are stored correctly. See e.g. `CohortSizeRange` class in `test-Rules-class.R` as a reference.

wwojciech (changes requested):
Looks very good. Just some minor comments.

R/Model-class.R
```diff
@@ -2653,6 +2653,92 @@ OneParExpNormalPrior <- function(skel_probs,
...
+# OneParExpExpPrior ----
+
+## class ----
+
+#' `OneParExpExpPrior`
+#'
+#' @description `r lifecycle::badge("experimental")`
+#'
+#' [`OneParExpExpPrior`] is the class for a standard CRM with an exponential prior on
+#' the log power parameter for the skeleton prior probabilities. It is an
```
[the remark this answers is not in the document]
danielinteractive: Hm I think the description can be very similar / analogous to the other `OneParExpNormalPrior` class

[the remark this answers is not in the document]
wwojciech: > Hm I think the description can be very similar / analogous to the other `OneParExpNormalPrior` class

Actually it is also misleading in `OneParExpNormalPrior`...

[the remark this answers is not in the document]
0liver0815: I think @wwojciech is right. It's misleading (its was copied and pasted from `OneParExpNormalPrior`) and should be corrected to
```
.... is the class for a standard CRM with an exponential prior on
power parameter for the skeleton prior probabilities.
```
because no exp transformation for the power parameter is used here.

R/Model-methods.R
0liver0815: ok

R/Model-class.R
0liver0815: will remove both

R/Model-class.R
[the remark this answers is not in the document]
danielinteractive: ok then let's fix both and make it consistent, thanks @0liver0815

R/Model-methods.R
0liver0815: @wwojciech That is a potential problem. Theta is sampled from the mcmc. If we check here, it ends with an error, but the user has no chance to change. I am wondering if jags returns 0 for theta.

tests/testthat/test-Model-class.R
0liver0815: @wwojciech I think I do not fully understand the logic. I tried the code
```
test_that(".OneParExpExpPrior works as expected", {
  result <- expect_silent(.OneParExpExpPrior())
  expect_valid(result, "OneParExpExpPrior")
})
```
But that end in an error, even though the class of result is `OneParExpExpPrior`.

R/Model-methods.R
wwojciech: I think it can return `0`, but the probability of that is equal `0`. Still, I would need add this assertion - and the function will fail, which is ok. @danielinteractive  - what do you think?

tests/testthat/test-Model-class.R
wwojciech: See for example this one:
```
test_that("CohortSizeRange object can be created with user constructor", {
  result <- expect_silent(CohortSizeRange(c(0, 30, 50), c(20, 60, 90)))
  expect_valid(result, "CohortSizeRange")
  expect_identical(result@intervals, c(0, 30, 50))
  expect_identical(result@cohort_size, c(20L, 60L, 90L))
})
```

R/Model-class.R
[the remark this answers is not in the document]
wwojciech: With regard to this, I think the names for these classes are wrong. It should be: `OneParLogNormalPrior` (instead of `OneParExpNormalPrior`) and `OneParExpPrior` (instead of `OneParExpExpPrior`).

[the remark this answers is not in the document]
0liver0815: > 2. The `OneParExpExpPrior` is wrong, it should `OneParExpPrior`, as the `theta` parameter in this model follows `Exp` distr.

But there are other 1-parameter models. The part `OneParExp` should highlight that it is the 1 parameter power model and the second part which prior is used `ExpPrior.`

> 1. The `OneParExpNormalPrior` class name is a bit confusing. Better would be: `OneParLogNormalPrior`, as the `theta` parameter in this model follows `LogNormal` distr. I mean, the current naming is still acceptable, because `exp(X) ~ LogNormal`, when `X ~ Normal`, but I just think `OneParLogNormalPrior` would be better.

I agree. It is just that I remember that I see often in literature that the normal distribution is mentioned for the power parameter and then a truncation or exp transformation is used to make it positive. I guess therefore the nomenclature was `OneParExpNormalPrior`.

R/Model-methods.R
0liver0815: I agree that it fails if it is 0. Could we probably add some tiny value to make the calculation run and issue a warning?

R/Model-class.R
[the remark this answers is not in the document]
wwojciech: > > 2. The `OneParExpExpPrior` is wrong, it should `OneParExpPrior`, as the `theta` parameter in this model follows `Exp` distr.
> 
> But there are other 1-parameter models. The part `OneParExp` should highlight that it is the 1 parameter power model and the second part which prior is used `ExpPrior.`
> 
> > 1. The `OneParExpNormalPrior` class name is a bit confusing. Better would be: `OneParLogNormalPrior`, as the `theta` parameter in this model follows `LogNormal` distr. I mean, the current naming is still acceptable, because `exp(X) ~ LogNormal`, when `X ~ Normal`, but I just think `OneParLogNormalPrior` would be better.
> 
> I agree. It is just that I remember that I see often in literature that the normal distribution is mentioned for the power parameter and then a truncation or exp transformation is used to make it positive. I guess therefore the nomenclature was `OneParExpNormalPrior`.

I see, thanks @0liver0815. Then `OneParExpNormalPrior` should be `OneParExpLogNormalPrior` or `OneParExpExpNormalPrior`, and `OneParExpExpPrior` is OK.

tests/testthat/test-Model-class.R
0liver0815: added

0liver0815: @wwojciech - Thanks. I think I got it now.

R/Model-methods.R
```diff
@@ -676,7 +676,7 @@ setMethod(
 ## OneParExpNormalPrior ----
 
 #' @describeIn dose compute the dose level reaching a specific target
-#'   probability of the occurrence of a DLE (`x`).
+#'   probability of the occurrence of a DLT (`x`).
```
wwojciech, line 679: Here should be just `probability of the occurrence of a DLT`

tests/testthat/test-Model-class.R
```diff
@@ -1094,6 +1094,122 @@ test_that("No NA is returned in dose calculations for OneParExpNormalPrior model
...
+  result <- expect_silent(
+    OneParExpExpPrior(
+      skel_probs = seq(from = 0.1, to = 0.9, length = 5),
+      dose_grid = 1:5,
+      lambda = 2
+    )
+  )
+  expect_valid(result, "OneParExpExpPrior")
+  expect_identical(result@skel_probs, seq(from = 0.1, to = 0.9, length = 5))
+  expect_identical(result@lambda, 2)
```
wwojciech, line 1111: How about checking `skel_fun` and `skel_fun_inv` slots?

examples/Model-class-OneParExpExpPrior.R
```diff
@@ -0,0 +1,5 @@
+my_model <- OneParExpExpPrior(
+  skel_probs = seq(from = 0.1, to = 0.9, length = 5),
```
wwojciech, line 2: Maybe better to manualy create these values rather than using `seq`, so `c(0.1, 0.3, 0.5, 0.7, 0.9)`.

tests/testthat/test-Model-class.R
```diff
@@ -1094,6 +1094,122 @@ test_that("No NA is returned in dose calculations for OneParExpNormalPrior model
...
 })
 
+# OneParExpExpPrior ----
+
+## constructor ----
+
+test_that("OneParExpExpPrior object can be created with user constructor", {
+  result <- expect_silent(
+    OneParExpExpPrior(
+      skel_probs = seq(from = 0.1, to = 0.9, length = 5),
```
wwojciech, line 1104: same here, better to create a vector manually rather than using `seq` fun.

wwojciech (changes requested):
Few minor comments remained.

R/Model-class.R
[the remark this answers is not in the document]
danielinteractive: hm I think no need to be too precise here. Let's stay with reasonable name choices even if they are not mathematically 100% precise

R/Model-methods.R
danielinteractive: but when `theta` is from real samples, it will never be exactly 0, so is this a real issue?

wwojciech: yes, `theta` will be a posterior samples, and then I think it will never be exactly `0`, but since we have this `doseFunction` concept which allows for the end user to supply any value for `thet`, including `0`, and even a negative ones, so I would still add the assertion for `theta` here.

R/Model-class.R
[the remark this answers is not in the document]
wwojciech: Agree, but at least I think we should be consistent, because at the moment `Exp` denotes exponential transformation (for `OneParExpNormalPrior`) and at the same time, the `Exp` denotes the distribution for `theta` (for `OneParExpExpPrior `).

R/Model-methods.R
danielinteractive: ok good the user input could really be 0. then yeah I would agree to add assertion. we won't get exact 0 from MCMC

0liver0815: Makes sense to me. I did not consider that the user may use the function to calculate dose or prob by himsellf. Will add the check.

R/Model-class.R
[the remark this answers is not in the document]
wwojciech: @0liver0815 After discussion with @danielinteractive, we agreed on this:
(old model): OneParLogNormalPrior: f(x) ^ exp(theta), theta ~ Normal()
(new model): OneParExpPrior: f(x) ^ theta, theta ~ Exp()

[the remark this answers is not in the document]
0liver0815: @wwojciech fine with me. I will rename the class accordingly.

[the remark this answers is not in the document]
wwojciech: Thanks @0liver0815 - just your class. I will replace the old one with a separate issue.

tests/testthat/test-Model-class.R
wwojciech: I mean to check the content of both functions. Yes, it is necessarily, as we must ensure, that the user constructor works in a way that is expected, that is really creates valid interpolation based on the input (i.e. `skel_probs`, `dose_grid`).

0liver0815:
Hi @wwojciech, all changes are made and checks are now completed without error. You can now finally review.

wwojciech (changes requested):
Thanks @0liver0815 It looks very good! I also added some minor changes so it is almost ready. But, I have noticed that there is no unit test for class validation function `v_model_one_par_exp_prior`... Did you create this?

0liver0815:
> Thanks @0liver0815 It looks very good! I also added some minor changes so it is almost ready. But, I have noticed that there is no unit test for class validation function `v_model_one_par_exp_prior`... Did you create this?

@wwojciech I missed that part. Sorry. Just added.

wwojciech (approved):
Perfect! Thanks a lot @0liver0815

wwojciech:
Hi @danielinteractive - I reviewed this PR as we discussed.

danielinteractive (approved):
thanks a lot @0liver0815 and @wwojciech for review! very nice

PR #450, opened 2022-12-06
Closes #439

### LINKED ISSUE #439: `probFunction`s should be vectorised
A function obtained from `probFunction()` throws an exception when its `dose` parameter is passed a vector.  This causes inefficiency when used in, for example, a `data.frame` or tibble.  For example

```
postSamples5 <- mcmc(
  data = fifthFullCohort,
  model = model,
  options = vignetteMcmcOptions
)

fullSamples <- tibble(
                 Alpha = postSamples5@data$alpha0,
                 Beta = postSamples5@data$alpha1
               ) %>%
               expand(nesting(Alpha, Beta), Dose = doseGrid)
```
followed by
```
fullSamples %>% mutate(P=probFunction(model, alpha0=Alpha, alpha1=Beta)(dose=Dose))
```
produces
```
Error in `mutate_cols()`:
! Problem with `mutate()` column `P`.
ℹ `P = probFunction(model, alpha0 = Alpha, alpha1 = Beta)(dose = Dose)`.
✖ Assertion on 'dose' failed: Must have length 1, but has length 3888.
Caused by error in `.local()`:
! Assertion on 'dose' failed: Must have length 1, but has length 3888.
```

Whereas
```
fullSamples %>% 
  rowwise() %>% 
  mutate(P=probFunction(model, alpha0=Alpha, alpha1=Beta)(dose=Dose)) %>% 
  ungroup()
# A tibble: 3,888 × 4
   Alpha  Beta  Dose         P
   <dbl> <dbl> <dbl>     <dbl>
 1 -2.44 1.93      1 0.0000366
 2 -2.44 1.93      3 0.000305 
 3 -2.44 1.93      9 0.00255  
 4 -2.44 1.93     20 0.0118   
 5 -2.44 1.93     30 0.0255   
 6 -2.44 1.93     45 0.0541   
 7 -2.44 1.93     60 0.0907   
 8 -2.44 1.93     80 0.148    
 9 -2.44 1.93    100 0.211    
10 -2.20 0.584     1 0.0104   
# … with 3,878 more rows
```
works, but is hugely inefficient.

wwojciech:
Draft PR - no unit tests yet, some existing test fail, which is ok.


wwojciech:
Hi @danielinteractive - I am submitting this draft PR (before any changes, addition to the unit tests) to confirm if this looks reasonable. I have some concern with repeating the following schema:
```
nsamples <- sampleSize(samples@options)
if (nsamples > 1L) {
  assert_length(x, len = c(1L, nsamples))
}
```
Wouldn't be better if we have e.g. this `if` clause inside the assertion, so for example we may only have the following assertion that check whether `x` is of compatible length with respect to `len`), i.e.:
```
assert_compatible_lengts(x = x, len = nsamples)
```

R/Model-methods.R
```diff
@@ -285,12 +286,16 @@ setMethod(
     samples = "Samples"
   ),
   definition = function(x, model, samples) {
+    assert_probabilities(x)
     assert_subset(c("alpha0", "alpha1"), names(samples@data))
+    nsamples <- sampleSize(samples@options)
+    if (nsamples > 1L) {
+      assert_length(x, len = c(1L, nsamples))
+    }
```
danielinteractive, line 294: could write in one line to save some real estate
```suggestion
    if (nsamples > 1L) assert_length(x, len = c(1L, nsamples))
```

danielinteractive (commented):
thanks @wwojciech , good start - see inside minor comments. regarding the Don't Repeat Yourself paradigm (DRY), yeah it is borderline I would say. (before we didn't have any assertions on the dimensions it seems) If you want to reduce the 2 lines to 1 line I would go for a very specific assertion function, like `assert_samples_doses()` or something like that. could also just be a simple helper function instead of a checkmate extension. but yeah, because it is just 2 to 1 line, borderline I would say. also ok as is (just compressing the if statment in 1 line here looks nicer)

danielinteractive:
@wwojciech name is ok , just make it specific to our use case for the functions, and not generic for all possible input length comparisons etc.

wwojciech:
> @wwojciech name is ok , just make it specific to our use case for the functions, and not generic for all possible input length comparisons etc.

Sure, thanks @wwojciech, Actually I found this `?vector_recycling_rules`, so I think the best name would be `assert_common_size`.

wwojciech:
Ok, @danielinteractive I made another version. Please let me know if this is ok, so that I will proceed with the unit tests. I am sorry for double-checking but there will be a lot of changes to the unit tests so it is better to confirm before I start to make any modifications. Tx!

R/Model-methods.R
```diff
@@ -637,13 +637,13 @@ setMethod(
...
+    assert_numeric(x, min.len = 1L, any.missing = FALSE, finite = TRUE)
     theta1 <- model@theta1
     theta2 <- model@theta2
     constant <- model@const
-    assert_length(theta1, 1L)
-    assert_length(theta2, 1L)
-    assert_length(constant, 1L)
+    assert_true(length(theta1) == 1L)
+    assert_true(length(theta2) == 1L)
+    assert_true(length(constant) == 1L)
```
danielinteractive, line 646: I think `assert_scalar` should be simpler here

danielinteractive (commented):
Thanks @wwojciech , do we need this `y_len` argument? Instead using an `assert_length(x, len)` would be better I think.

R/Model-methods.R
```diff
@@ -1433,6 +1458,7 @@ setMethod(
   definition = function(dose, model_dle, samples_dle, model_eff, samples_eff, ...) {
     dle <- prob(dose, model_dle, samples_dle)
     eff <- efficacy(dose, model_eff, samples_eff)
+    assert_common_length(dle, eff)
```
wwojciech, line 1461: #450 Actually we need `y` argument here

wwojciech:
> Thanks @wwojciech , do we need this `y_len` argument? Instead using an `assert_length(x, len)` would be better I think.

Yes, we need it here: https://github.com/Roche/crmPack/blob/45517ceb4818d62db7808ec20ebc397e9bb7bfab/R/Model-methods.R#L1461 and also here https://github.com/Roche/crmPack/blob/45517ceb4818d62db7808ec20ebc397e9bb7bfab/R/Model-methods.R#L1487

danielinteractive (approved):

Issue #449, opened 2022-12-06
Create new `sampleSize` method for `Samples` class


wwojciech:
It should return the size of the sample, e.g.: 
```
my_data <- Data(
  x = c(0.1, 0.5, 1.5, 3, 6, 10, 10, 10),
  y = c(0, 0, 0, 0, 0, 0, 1, 0),
  cohort = c(0, 1, 2, 3, 4, 5, 5, 5),
  doseGrid = c(0.1, 0.5, 1.5, 3, 6, seq(from = 10, to = 80, by = 2))
)

# Initialize a model, e.g. 'LogisticLogNormal'.
my_model <- LogisticLogNormal(
  mean = c(-0.85, 1),
  cov = matrix(c(1, -0.5, -0.5, 1), nrow = 2),
  ref_dose = 56
)

# Get samples from posterior.
my_options <- McmcOptions(burnin = 100, step = 2, samples = 20)
my_samples <- mcmc(data = my_data, model = my_model, options = my_options)
```
New function:
```
> sampleSize(my_samples)
[1] 20
```
Note that the generic function `sampleSize` is already created in `McmcOptions`.

Tasks:

1. Create new method
2. Replace existing `sampleSize(samples@options)` with `sampleSize(samples)`, wherever possible.

PR #463, opened 2022-12-22
Closes #154

### LINKED ISSUE #154: doseFunction-GeneralModel and probFunction-GeneralModel - handle params when array
In doseFunction-GeneralModel (and probFunction-GeneralModel):

```
samples <- Samples(
    data = model_params,
    options = McmcOptions(samples = length(model_params[[1]]))
)
```
model_params might ba also an array object, so length is not always appropriate. This should be corrected. See `h_null_if_scalar()` fun on how this can be handled.

wwojciech:
Closes #154


danielinteractive:
@wwojciech btw the "Assignee" for PRs is the reviewer, not the author :-)

NEWS.md
```diff
@@ -1,4 +1,5 @@
 # Version 1.0.9000.9122
+* Removed helper `h_null_if_scalar`.
```
danielinteractive, line 2: not needed for NEWS since internal change
```suggestion
```

wwojciech:
Using this issue, I removed also `h_null_if_scalar` function which is not needed anymore (due to https://github.com/Roche/crmPack/issues/449)

tests/testthat/test-Model-methods.R
```diff
@@ -21,6 +21,30 @@ test_that("doseFunction-GeneralModel returns correct dose function", {
...
+  )
+  samples <- h_as_samples(params, burnin = 10000, fixed = FALSE)
+  dose_args <- c("x", "model", "samples")
+
+  dose_fun <- doseFunction(model, alpha0 = params$alpha0, alpha1 = params$alpha1, comp = params$comp)
+  dose_fun_dose_args <- as.character(body(dose_fun)[[2]][-1])
+  dose_fun_env <- environment(dose_fun)
+
+  expect_function(dose_fun, args = "x", nargs = 1, null.ok = FALSE)
+  expect_equal(dose_fun_dose_args, dose_args)
```
danielinteractive, line 39: `dose_args` is only used here - better readable if you avoid that variable and just put the character vector here
```suggestion
  expect_equal(dose_fun_dose_args, c("x", "model", "samples"))
```

tests/testthat/test-Model-methods.R
```diff
@@ -21,6 +21,30 @@ test_that("doseFunction-GeneralModel returns correct dose function", {
...
+  params <- list(
+    alpha0 = matrix(c(-0.94, -0.94, -2.37, -2.37, -0.67, -0.67, -1.28, -1.08), nrow = 4),
+    alpha1 = matrix(c(0.45, 0.45, 0.40, 0.40, 0.75, 0.75, 1.18, 0.63), nrow = 4),
+    comp = c(1, 1, 1, 1)
+  )
+  samples <- h_as_samples(params, burnin = 10000, fixed = FALSE)
+  dose_args <- c("x", "model", "samples")
+
+  dose_fun <- doseFunction(model, alpha0 = params$alpha0, alpha1 = params$alpha1, comp = params$comp)
+  dose_fun_dose_args <- as.character(body(dose_fun)[[2]][-1])
```
danielinteractive, line 35: this is too complicated for me - what is this? please add comment and check if cannot be simplified

tests/testthat/test-Model-methods.R
```diff
@@ -21,6 +21,30 @@ test_that("doseFunction-GeneralModel returns correct dose function", {
...
+  dose_fun_env <- environment(dose_fun)
+
+  expect_function(dose_fun, args = "x", nargs = 1, null.ok = FALSE)
+  expect_equal(dose_fun_dose_args, dose_args)
+  expect_subset(
+    setdiff(dose_fun_dose_args, "x"),
```
danielinteractive, line 41: better put explicit character vector here

tests/testthat/test-Model-methods.R
```diff
@@ -87,6 +111,30 @@ test_that("probFunction-GeneralModel returns correct prob function", {
...
+  params <- list(
+    alpha0 = matrix(c(-0.94, -0.94, -2.37, -2.37, -0.67, -0.67, -1.28, -1.08), nrow = 4),
+    alpha1 = matrix(c(0.45, 0.45, 0.40, 0.40, 0.75, 0.75, 1.18, 0.63), nrow = 4),
+    comp = c(1, 1, 1, 1)
+  )
+  samples <- h_as_samples(params, burnin = 10000, fixed = FALSE)
+  prob_args <- c("dose", "model", "samples")
+
+  prob_fun <- probFunction(model, alpha0 = params$alpha0, alpha1 = params$alpha1, comp = params$comp)
+  prob_fun_prob_args <- as.character(body(prob_fun)[[2]][-1])
```
danielinteractive, line 125: similar comments apply here

danielinteractive (commented):
thanks @wwojciech , please see inside - let's try to make the tests as readable as possible, thanks

tests/testthat/test-Model-methods.R
wwojciech: This is:
```
> dose_fun
function(x) {
      dose(x, model, samples)
    }
<environment: 0x7fd861415190>
> dose_fun_dose_args <- as.character(body(dose_fun)[[2]][-1])
> dose_fun_dose_args
[1] "x"       "model"   "samples"
```
so with `dose_fun_dose_args <- as.character(body(dose_fun)[[2]][-1])` I am trying to get the argument names from the inner dose function.

danielinteractive: Aha yeah please add comments

wwojciech: Hi @danielinteractive I slightly improved the code (add more robust and flexible checks) and also added comments. I did it only for `doseFunction-GeneralModel returns correct dose function` test. Please let me know if this looks ok so that I will propagate it to other tests.

wwojciech:
Hi @cicdguy do you know why `R CMD Check` does not start here? Tx!

tests/testthat/test-Model-methods.R
```diff
@@ -4,21 +4,72 @@
...
   model <- h_get_logistic_log_normal()
-  samples <- h_as_samples(list(alpha0 = 1, alpha1 = 2), burnin = 10000, fixed = FALSE)
-  dose_args <- c("x", "model", "samples")
+  samples <- Samples(list(alpha0 = 1, alpha1 = 2), options = McmcOptions(samples = 1))
 
   dose_fun <- doseFunction(model, alpha0 = 1, alpha1 = 2)
-  dose_fun_dose_args <- as.character(body(dose_fun)[[2]][-1])
   dose_fun_env <- environment(dose_fun)
 
+  expect_equal(formalArgs(doseFunction), c("model", "..."))
```
danielinteractive, line 12: This can do a bit easier with the `expect_function()` from checkmate

danielinteractive: Or did you try that already ?

danielinteractive (approved):

tests/testthat/test-Model-methods.R
wwojciech: Actually, I think there is no need to check is the `doseFunction` is a function, I mean - this is the main function we want to test to there is no way `doseFunction` is not a function? It is like for instance `dose` function - we do not test if this is a function, but rather we test its results. What do you think @danielinteractive ?

danielinteractive: ah my point is that with that expectation function you can check the arguments a bit easier

wwojciech: Ah, sure - I will do that then. Tx!

PR #466, opened 2023-01-03
Closes #196

### LINKED ISSUE #196: `getEff()` method - improvement
After I more learned about the usage of `getEff()` method, I think it would be great to introduce a new argument to this function, lets say called, `no_dlt`, which would be a flag that instructs whether only DLT elements should be returned, or not.

@danielinteractive - FYI

wwojciech:
Closes #196

### DIFF
--- a/R/Data-methods.R
+++ b/R/Data-methods.R
@@ -512,7 +512,7 @@ setMethod(
 #'
 #' @param object (`DataDual`)\cr object from which the responses and dose levels
 #'   are extracted.
-#' @param ... not used.
+#' @param ... further arguments passed to class-specific methods.
 #' @return `list` with efficacy responses categorized by the DLT value.
 #' @export
 #'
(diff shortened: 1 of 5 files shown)


R/Data-methods.R
```diff
@@ -527,33 +527,36 @@ setGeneric(
...
-      x_no_dlt = x_no_dlt,
-      w_no_dlt = w_no_dlt
-    )
+    if (!no_dlt) {
+      eff_dlt <- if (any(is_dlt)) {
+        list(x_dlt = object@x[is_dlt], w_dlt = object@w[is_dlt])
+      } else {
+        list(x_dlt = NULL, w_dlt = NULL)
+      }
+      eff <- c(eff, eff_dlt)
```
danielinteractive, line 558: probably can be simplified - if there are no dlts then `eff` is not changed

danielinteractive (commented):
thanks Wojciech!

R/Data-methods.R
```diff
@@ -527,33 +527,36 @@ setGeneric(
...
-
-    is_dlt <- object@y == 1
-    if (any(is_dlt)) {
-      x_dlt <- object@x[is_dlt]
-      w_dlt <- object@w[is_dlt]
-    }
-    if (any(!is_dlt)) {
-      x_no_dlt <- object@x[!is_dlt]
-      w_no_dlt <- object@w[!is_dlt]
+  definition = function(object, no_dlt = FALSE) {
```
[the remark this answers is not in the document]
wwojciech: I think for methods we do not need it. For example for `dose` and `prob` methods we do not have `...` args.

R/Data-methods.R
wwojciech: In case no dlt events (i.e. `any(is_dlt)` is `FALSE`), then we still need `list(x_dlt = NULL, w_dlt = NULL)`. But, if the user wants to see only no dlt responses (i.e. `no_dlt` flag is `TRUE`), then `x_dlt` and `w_dlt` slots must not appear in the output list.

danielinteractive (approved):

PR #471, opened 2023-01-12
Closes #467

### LINKED ISSUE #467: `efficacy-EffFlexi` method improvement
1. vectorised `dose` argument handling. Three cases should be possible:
1.1. 1 dose and >1 sample
1.2 >1 dose + 1 sample 
1.3. n doses + n samples 
2. better assertions.
3. NA returned for doses outside of the dose grid range (and warning thrown).
4. assert that the `dose_grid` is of length at least 1.
5. placebo (if any) to be included in the interpolation.
6. add new unit tests accordingly.

wwojciech:
Closes #467


NAMESPACE
```diff
@@ -455,6 +455,7 @@ importFrom(parallel,makeCluster)
 importFrom(parallel,stopCluster)
 importFrom(rjags,jags.model)
 importFrom(rjags,jags.samples)
+importFrom(stats,approx)
```
danielinteractive, line 458: Please let's use `stats::approx` and not import it , same for others

danielinteractive (approved):

PR #485, opened 2023-02-01
Remove CPU limit

0liver0815:
@danielinteractive I did this very small bugfix, as I wanted to use more cpu with the current package.


danielinteractive:
Actually one simple idea; just use instead `parallelly::availableCores()` that is safer

danielinteractive:
Probably need to put parallelly into Imports in the DESCRIPTION file

danielinteractive (approved):

PR #490, opened 2023-02-20
124 report stopping

### LINKED ISSUE #94: Add additional statistics to the analysis of simulation output
For the final protocol we analyze and present statistics that are currently not covered by summary method in crmPack. In addition, the summary method in crmPack is tailored for the NextBestNCRM-class but fit not very well for other next best dose approaches, e.g. smallest distance to target we want to add to crmPack.

Here is an example what we have presented in past in the protocol:

image

In the future we will change this reporting, e.g. we will add the selection probability and probably exchange min/max by the 10% and 90% quantiles as currently presented in crmPack. However, for assessment of the operation characteristics in general it will still be useful to look also at min/max. 

Some of the statistics that are presented cannot be derived from the current S4 object of class simulation in crmPack (e.g. median MTD) or are not easy to derive (e.g. reason for stopping).
 
Therefore, changes to existing functions in crmPack may be necessary. In the following we want to give more details using the stopping rules as an example.

In principle, the current S4 object of class simulation in crmPack contains the stopReasons as list. The individual results why a study has stopped, is captured as text and in one can identify which rule is triggered. 
As the stopping text contains detailed information that vary between simulated studies and in case the rule is triggered or not, it is not straight forward to be used for further reporting.

We tried the following ways to make the name of the stopping rule and the status of a stopping rule available to the simulation output object:
1)	Via the text that is returned by the stopping rule:
A slot ‘label’ is added to the stopping rule. The label (stopping@label) and  the value of the variable doStop are added to the returned text, so that it looks like
… paste(doStop, ' : ', stopping@label, ' : ', …text as before….)
2)	Via the simulation function:
The simulation function is updated so that the name of the stopping rule and the status of the stopping rule is written to the simulation output.

The advantage of implementation 1 is that a label is introduced that can be directly used for reporting. The disadvantage is that the output of any stopping rule must be changed to follow the new output text pattern so that it can be analyzed automatically.

The advantage of implementation 2 is that the individual stopping rules do not need to be adapted. A possible disadvantage is that the simulation function needs to be adapted.

Given the goal that additional statistics can be derived from the simulation output (e.g. median MTD) a change of the simulation function seems to be necessary in any case.

An Idea of implantation without affecting the current reporting mechanism, could be to create an additional (new) design class, that is picked up by a revised simulation function and reporting function. This may be needed anyway, when another mechanism for selecting the next best dose is used (no interval around the target toxicity is used).

We like to get thought if this makes sense at all and if so, how to implement this in the best way into the package. Probably there better ways to archive the reporting goal.

ClaraBeckBayer:
# Pull Request

Fixes #94


danielinteractive:
Thanks @ClaraBeckBayer , @wwojciech could you maybe have a first review look at this? I think we discussed this a bit in the meetings before so you probably also have some context by now.

wwojciech:
> Thanks @ClaraBeckBayer , @wwojciech could you maybe have a first review look at this? I think we discussed this a bit in the meetings before so you probably also have some context by now.

Sure @danielinteractive, especially that I have been already working on this a bit with @ClaraBeckBayer

wwojciech:
@ClaraBeckBayer - which issue is suppose to be closed with this PR? Tx!

danielinteractive:
I think #94 as mentioned in description? Basically the reporting of stopping rule results

danielinteractive:
@wwojciech Would you have time to start the review?

wwojciech:
> @wwojciech Would you have time to start the review?

Hi @danielinteractive - I started it, just need to finish it :)

R/Simulations-class.R
```diff
@@ -91,54 +108,72 @@ GeneralSimulations <- function(data,
...
+    prototype(
+      fit =
+        list(
+          c(0.1, 0.2),
+          c(0.1, 0.2)
+        ),
+      stopReasons =
+        list("A", "A"),
+      stop_report =
+        matrix(TRUE, TRUE)
```
wwojciech, line 133: is it reasonable init for `stop_report`? It is the same as `stop_report = matrix(TRUE)`?

R/Simulations-class.R
```diff
@@ -91,54 +108,72 @@ GeneralSimulations <- function(data,
...
+        nSims <- length(object@data)
+
+        o$check(
+          identical(length(object@fit), nSims),
+          "fit must have same length as data"
+        )
+        o$check(
+          identical(length(object@stopReasons), nSims),
+          "stopReasons must have same length as data"
+        )
```
wwojciech, line 149: don't we need validation for `stop_report` slot?

wwojciech (changes requested):
Hi @robertadamsbayer @ClaraBeckBayer - I started to review this PR sometimes ago and I am continuing now. However, after several hours I spent on this, I need to stop at this moment and get this PR back to you. The problem here is that you reformatted hundreds of lines of the code which is not relevant to this issue but it was just styled with styler (I guess bu mistake?). This makes a trustworthy review almost impractical. Look for example for `Simulations-methods.R` file. Almost every line in this file has changed and the file itself has almost 2500 lines of code.
Would it be possible you only apply the changes that are relevant to the PR without making any other changes? Also, you may have a look at the review questions/doubts that I have already raised. Thanks a lot and please let me know if there is anything I could clarify more or help. We can also have a quick chat on Slack if needed.

danielinteractive:
@wwojciech I think this might just be whitespace changes. Please try to disable the show of whitespace changes in the Files Changed view. Then it looks much less. I just checked the file you mentioned and only few lines are changed.

wwojciech:
> @wwojciech I think this might just be whitespace changes. Please try to disable the show of whitespace changes in the Files Changed view. Then it looks much less. I just checked the file you mentioned and only few lines are changed.

Thanks @danielinteractive - it was white spaces but also a lot of new lines. I styled the whole package on Friday so now this problem is gone.

R/Rules-methods.R
```diff
@@ -1647,7 +1636,8 @@ setMethod("stopTrial",
       overallText <- lapply(individualResults, attr, "message")
 
       return(structure(overallResult,
-        message = overallText
+        message = overallText,
+        individual = individualResults
       ))
```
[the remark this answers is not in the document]
ClaraBeckBayer: Hi Wojciech, 
ths is for StoppingList, right? 
From our discussion before I understood that "StoppingList" is an outdated version of combining stopping rules and will eventually be removed. This is why we focused on StoppingAll and StoppingAny only.

R/Rules-methods.R
```diff
@@ -1810,7 +1816,8 @@ setMethod("stopTrial",
```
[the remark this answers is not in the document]
ClaraBeckBayer: Those are two different things. report_label is the exact labels as it is returned for reporting later on (idea is: that will be displayed in the reporting tables later).

R/Simulations-methods.R
```diff
@@ -528,6 +528,12 @@ setMethod("summary",
       propAtTarget <- mean((toxAtDoses > target[1]) &
         (toxAtDoses < target[2]))
 
+      if (.hasSlot(object, "stop_report") == TRUE) {
```
wwojciech, line 531: I think that every object of class `Simulations` should have this slot. Values between the instances may differ but every instance should have this slot, I guess.

R/Simulations-methods.R
```diff
@@ -839,6 +844,24 @@ setMethod("show",
...
 
+      # report stopping rules
+      logPercent <- apply(object@stop_report, 2, mean) * 100
+
+
+
+      if (!is.null(names(logPercent))) {
+        cat("Stopping rules:\n \n")
+        for (i in 1:ncol(object@stop_report)) {
+          if (!is.na(names(logPercent)[i])) {
```
wwojciech, line 855: this can be written more compact. Maybe first determine the indices for which the name is not NA. Also I do not see how it may happen that we have `NA` names.....?

robertadamsbayer: Updated - the logic is now: If there is ANY column containing no <NA> the Stopping rule reporting is initiated. For all non-<NA> columns the label and the logPercentage is output. The presence of <NA> or non-<NA> is the switch to output Stopping Rule reporting.

wwojciech: Ok, then this code is more compact I think:
```
# Report individual stopping rules with non-<NA> labels.
stop_pct <- colMeans(object@stop_report) * 100
stop_pct_to_print <- stop_pct[!is.na(names(stop_pct))]
if (length(stop_pct_to_print) > 0) {
  cat(
    "Stopping rules:\n \n",
    paste(names(stop_pct_to_print), ": ", stop_pct_to_print, "%\n \n")
  )
}
```

wwojciech:
@ClaraBeckBayer , @robertadamsbayer - I see there are some errors in the code, so that the package cannot be build.

wwojciech:
Why is this unit test: `"reporting labels and logicals are correctly returned using the recursive function in simulate"` in `test-Design-methods.R` empty?

wwojciech:
> @ClaraBeckBayer, @robertadamsbayer - I see there are some errors in the code, so that the package cannot be build.

Ok, I solved it.

R/Simulations-methods.R
robertadamsbayer: Oh wow - that makes a lot of sense 💯 Thank you very much!

R/Simulations-class.R
```diff
@@ -108,65 +108,76 @@ GeneralSimulations <- function(data,
...
+    o$check(
+      !any(is.na(object@stop_report)),
+      "stop_report must not contain NA values"
+    )
+
+    o$check(
+      is.logical(object@stop_report),
+      "stop_report must be a matrix of class logical"
+    )
+
```
wwojciech, line 155: Combine these 3 above tests for `stop_report` into a single test using `checkmate::expect_matrix()`function.

wwojciech (changes requested):
The package does not pass automation check after your recent changes, please check it.

R/Rules-methods.R
```diff
@@ -1725,6 +1730,15 @@ setMethod("stopTrial",
       ## so can we stop?
       doStop <- nCohorts >= stopping@nCohorts
 
+      report_label <- if (isTRUE(stopping@report_label == "")) {
+        paste(
+          "\u2265", x@nCohorts, "cohorts in", x@percentage,
```
wwojciech, line 1735: what is the `x` here? I do not see this object on the list of explicit arguments to this function.

R/Rules-methods.R
```diff
@@ -1776,6 +1791,15 @@ setMethod("stopTrial",
       ## so can we stop?
       doStop <- nPatients >= stopping@nPatients
 
+      report_label <- if (isTRUE(stopping@report_label == "")) {
+        paste(
+          "\u2265", x@nPatients, "patients in",
+          x@percentage, "% dose range around NBD"
```
wwojciech, line 1797: same here: what is the `x` here? I do not see this object on the list of explicit arguments to this function.

R/Rules-methods.R
```diff
@@ -1973,6 +2023,15 @@ setMethod("stopTrial",
       ## so can we stop?
       doStop <- prob >= stopping@prob
 
+      report_label <- if (isTRUE(stopping@report_label == "")) {
+        paste(
+          "P(MTD >", x@thresh, "* NBD | P(DLE) = ", x@target, ") \u2265 ",
```
wwojciech, line 2028: same here: what is the `x` here? I do not see this object on the list of explicit arguments to this function.

R/Rules-methods.R
```diff
@@ -2030,6 +2090,12 @@ setMethod(
     mtd_cv <- (mad(mtd_samples) / median(mtd_samples)) * 100
     do_stop <- (mtd_cv <= stopping@thresh_cv) && (mtd_cv >= 0)
 
+    report_label <- if (isTRUE(stopping@report_label == "")) {
+      paste("P(MTD >", x@thresh, "* NBD | P(DLE) = ", x@target, ") \u2265 ", x@prob)
```
wwojciech, line 2094: same here: what is the `x` here? I do not see this object on the list of explicit arguments to this function.

R/Rules-methods.R
```diff
@@ -2183,6 +2261,16 @@ setMethod("stopTrial",
       ## so can we stop?
       doStop <- probTarget >= stopping@prob
 
+      report_label <- if (isTRUE(stopping@report_label == "")) {
+        paste(
+          "P(", x@target[1], "\u2264", "Biomarker \u2264", x@target[2],
```
wwojciech, line 2266: same here: what is the `x` here? I do not see this object on the list of explicit arguments to this function.

wwojciech: the content of the `paste()` can be in 1 line.

R/Rules-methods.R
```diff
@@ -2620,14 +2733,28 @@ setMethod("stopTrial",
...
+          paste(
+            "TD",
+            x@targetRatio,
+            "for",
+            x@targetEndOfTrial,
+            "target prob"
+          )
+      } else {
+        report_label <- stopping@report_label
+      }
```
wwojciech, line 2748: same here: why do we have this `stopping@report_label == "default"`, I think we changed already to "". Also `report_label` should be places out of `if-else` clause. Also I do not see `x` object here, so you cannot use `x@<slotname>`.

R/Rules-methods.R
```diff
@@ -2735,6 +2862,20 @@ setMethod("stopTrial",
...
+          paste(
+            "GStar",
+            x@targetRatio,
+            "for",
+            x@targetEndOfTrial,
+            "target prob"
+          )
+      } else {
+        report_label <- stopping@report_label
+      }
```
wwojciech, line 2877: same here: why do we have this `stopping@report_label == "default"`, I think we changed already to "". Also `report_label` should be places out of `if-else` clause. Also I do not see `x` object here, so you cannot use `x@<slotname>`.

R/Simulations-class.R
wwojciech: Just use `test_matrix()` here instead of all these `checkmate::testMatrix()`, `checkmate::testLogical()`. Let me know please if you are not sure about this.

examples/Stopping-rules-Reporting.R
```diff
@@ -0,0 +1,91 @@
...
+# Define the dose-grid
+emptydata <- Data(doseGrid = c(1, 3, 5, 10, 15, 20, 25, 40, 50, 80, 100))
+
+# Initialize the CRM model
+model <- LogisticLogNormal(
+  mean = c(-0.85, 1),
+  cov =
+    matrix(c(1, -0.5, -0.5, 1),
+      nrow = 2
+    ),
```
wwojciech, line 12:
```suggestion
  cov = matrix(c(1, -0.5, -0.5, 1), nrow = 2),
```

examples/Stopping-rules-Reporting.R
```diff
@@ -0,0 +1,91 @@
...
+  ref_dose = 56
+)
+
+# Choose the rule for selecting the next dose
```
wwojciech, line 16:
```suggestion
# Choose the rule for selecting the next dose.
```

examples/Stopping-rules-Reporting.R
```diff
@@ -0,0 +1,91 @@
...
+  nsim = 1,
+  seed = 819,
+  mcmcOptions = options,
+  parallel = FALSE
+))[3]
+
+# Summarize the Results of the Simulations
+summary(mySims, truth = myTruth)
+
+# nolint end
```
wwojciech, line 91: remove nolint from here.

wwojciech: Also the overall comment: we use snake_case style for variable names rather than camelCase. So, no `myTruth` by `my_truth `.

examples/Stopping-rules-Reporting.R
```diff
@@ -0,0 +1,91 @@
...
+  startingDose = 3
+)
+
+## define the true function
+myTruth <- probFunction(model, alpha0 = 7, alpha1 = 8)
+
+# Run the simulation on the desired design
+# We only generate 1 trial outcomes here for illustration, for the actual study
+# this should be increased of course
+options <- McmcOptions(
```
wwojciech, line 74:
```suggestion
my_options <- McmcOptions(
```

examples/Stopping-rules-Reporting.R
```diff
@@ -0,0 +1,91 @@
...
+)
+
+# Choose the rule for dose increments
+myIncrements <- IncrementsRelative(
+  intervals = c(0, 20),
+  increments = c(1, 0.33)
+)
+
+# Initialize the design
+design <- Design(
```
wwojciech, line 58:
```suggestion
my_design <- Design(
```

tests/testthat/test-Design-methods.R
```diff
@@ -92,3 +92,99 @@ test_that("NextBestInfTheory produces consistent results with a dataset", {
...
+test_that("reporting labels and logicals are correctly returned using the recursive function in simulate", {
+  my_data <- h_get_data(placebo = FALSE)
+  model <- LogisticLogNormal(
+    mean = c(-0.85, 1),
+    cov =
+      matrix(c(1, -0.5, -0.5, 1),
+        nrow = 2
+      ),
+    ref_dose = 56
+  )
```
wwojciech, line 105:
```suggestion
  model <- h_get_logistic_normal()
```
and also the same comment in other places.

tests/testthat/test-Design-methods.R
```diff
@@ -92,3 +92,99 @@ test_that("NextBestInfTheory produces consistent results with a dataset", {
...
+
+
+
+  increments <- IncrementsRelative(
+    intervals = c(0, 20),
+    increments = c(1, 0.33)
+  )
+  new_my_next_best <- NextBestNCRM(
```
wwojciech, line 113:
```suggestion
  new_next_best <- NextBestNCRM(
```

tests/testthat/test-Design-methods.R
```diff
@@ -92,3 +92,99 @@ test_that("NextBestInfTheory produces consistent results with a dataset", {
...
+    target = c(0.2, 0.35),
+    overdose = c(0.35, 1),
+    max_overdose_prob = 0.25
+  )
+  my_size <- CohortSizeConst(size = 3)
```
wwojciech, line 118:
```suggestion
  size <- CohortSizeConst(size = 3)
```

tests/testthat/test-Design-methods.R
```diff
@@ -92,3 +92,99 @@ test_that("NextBestInfTheory produces consistent results with a dataset", {
   expect_equal(result@toxAtDosesSelected, rep(1L, 5))
   expect_snapshot(result@meanFit)
 })
+
+test_that("reporting labels and logicals are correctly returned using the recursive function in simulate", {
+  my_data <- h_get_data(placebo = FALSE)
```
wwojciech, line 97:
```suggestion
  data <- h_get_data(placebo = FALSE)
```
In unit tests we do not use `my_` prefix.

tests/testthat/test-Design-methods.R
```diff
@@ -92,3 +92,99 @@ test_that("NextBestInfTheory produces consistent results with a dataset", {
...
+  new_my_next_best <- NextBestNCRM(
+    target = c(0.2, 0.35),
+    overdose = c(0.35, 1),
+    max_overdose_prob = 0.25
+  )
+  my_size <- CohortSizeConst(size = 3)
+
+
+
+  my_truth <- probFunction(model, alpha0 = 175, alpha1 = 5)
```
wwojciech, line 122:
```suggestion
  truth <- probFunction(model, alpha0 = 175, alpha1 = 5)
```

tests/testthat/test-Design-methods.R
```diff
@@ -92,3 +92,99 @@ test_that("NextBestInfTheory produces consistent results with a dataset", {
...
+  stopping <- StoppingAll(
+    stop_list =
+      list(
+        StoppingAny(
+          stop_list =
+            list(stopping1, stopping2),
+          report_label = "label_StoppingAny"
+        ),
+        stopping3
+      ),
```
wwojciech, line 143:
```suggestion
    stop_list = list(
        StoppingAny(stop_list = list(stopping1, stopping2), report_label = "label_StoppingAny"),
        stopping3
      ),
```

tests/testthat/test-Design-methods.R
```diff
@@ -92,3 +92,99 @@ test_that("NextBestInfTheory produces consistent results with a dataset", {
...
+  model <- LogisticLogNormal(
+    mean = c(-0.85, 1),
+    cov =
+      matrix(c(1, -0.5, -0.5, 1),
+        nrow = 2
+      ),
+    ref_dose = 56
+  )
+
+
```
wwojciech, line 107: remove these empty lines.

tests/testthat/test-Rules-methods.R
```diff
@@ -1453,7 +1465,55 @@ test_that("StoppingSpecificDose correclty replaces next best string with specifi
...
+    stopping = my_stopping,
+    model = my_model,
+    samples = my_samples,
+    dose = 7,
+    data = my_data
+  )
+
+
+  expect_equal(attr(result, "report_label"), "a customized label")
+})
```
wwojciech, line 1519: Ok, but we need to test it for each stopping class, not only for `StoppingMinCohorts`. So I would propose that we always test for a default label (within the existing unit tests) + we need a separate short test for custom label for each class.

wwojciech (changes requested):
Hi @ClaraBeckBayer , @robertadamsbayer Thanks a lot for all the changes you did. However, there is still something to correct. From the most important topic: we need to test against `report_label` for `stopTrial()` method for each stopping class, not only for `StoppingMinCohorts`. So I would propose that we always test for a default label (within the existing unit tests that are already in `test-Rules-methods.R` file) + we need a separate short test for custom label for each class. Let me know if there is anything unclear.

R/Rules-methods.R
```diff
@@ -1725,6 +1730,15 @@ setMethod("stopTrial",
       ## so can we stop?
       doStop <- nCohorts >= stopping@nCohorts
 
+      report_label <- if (isTRUE(stopping@report_label == "")) {
```
wwojciech, line 1733: I think this looks better:
```suggestion
      report_label <- if (isTRUE(nzchar(stopping@report_label))) {
```
and in all remaining places the same change

R/Rules-methods.R
```diff
@@ -1725,6 +1730,15 @@ setMethod("stopTrial",
...
+        paste(
+          "\u2265", x@nCohorts, "cohorts in", x@percentage,
+          "% dose range around NBD"
+        )
+      } else {
+        stopping@report_label
+      }
```
wwojciech, line 1740:
```suggestion
      report_label <- if (isTRUE(nzchar(stopping@report_label))) {
        stopping@report_label
      } else {
        paste(
          "\u2265", x@nCohorts, "cohorts in", x@percentage, "% dose range around NBD"
        )
      }
```

R/Simulations-class.R
wwojciech: This is not solved.

robertadamsbayer: Thanks for the hint - sorry - we missed that!

wwojciech:
Hi @ClaraBeckBayer , @robertadamsbayer - can you pl. do a `Pull` from the latest update that I made now, so that before you `Push` any changes you already have the latest changes? This is to avoid some issue in the way GitHub marks the changes relative to the main branch. Thanks a lot!

wwojciech:
@ClaraBeckBayer, @robertadamsbayer - can you also please mark all of these PR comments that are already addressed/solved. Tx!

R/Rules-methods.R
wwojciech: same here - not corrected yet.

tests/testthat/test-Rules-methods.R
wwojciech: this is still not solved.

wwojciech (changes requested):
Hi @ClaraBeckBayer @robertadamsbayer - there are still some changes to be applied before I can approve it. Please let me know in case of anything unclear. Tx!

wwojciech (changes requested):
Please apply the changes according to my last comment :).

robertadamsbayer:
Sorry for the delay working on the PR - some busy weeks... as usual. 

StoppingLowestDoseHSRBeta - no default yet (needs to be defined)

robertadamsbayer:
StoppingList with no report_label paramater. Alternativ test?

robertadamsbayer:
image

robertadamsbayer:
The warnings in the mentioned test is beyond our influence. It´s originated from the simulate() function: 

image

I think this is a generic warning that occurs in multiple places in crmPack due to ggplot2 functionality?

robertadamsbayer:
I might have found an issue but want to confirm first. While trying to solve the erroneous test function I realized that tweaking parameters leads to malformed stop_reports. This might not necessarily related to the report itself but to unexpected stoppings. We need to investiage that further? 

image

vs.

image

danielinteractive:
@robertadamsbayer @ClaraBeckBayer I guess this can be closed?

PR #491, opened 2023-03-13
Test sas results

marlenesg:
# Pull Request

incorporating test to compare crmPack results with results computed by SAS


tests/testthat/test-sas-results-part-1.R
```diff
@@ -0,0 +1,139 @@
...
+    burnin = 5000,
+    step = 2,
+    samples = 1000000,
+    rng_kind = "Wichmann-Hill",
+    rng_seed = 1
+  )
+
+  dose_grid_sc1 <- c(10, 20, 35, 50, 65, 80, 90, 100)
+  tab_sc1 <- vector("list")
+  tab_sc1$"Data example 1 (DLTs: 0)" <- Data(
```
danielinteractive, line 16: For maintenance and readability, it would actually be better to have each data example in a separate `test_that()` call. Would that be possible?

tests/testthat/test-sas-results-part-1.R
```diff
@@ -0,0 +1,398 @@
...
+    POVEREX = dose_rec_loss$probs[, "excessive"],
+    POVERUN = dose_rec_loss$probs[, "unacceptable"],
+    POVER = rowSums(dose_rec_loss$probs[, c("excessive", "unacceptable")]),
+    PMEAN = dose_rec_loss$probs[, "mean"],
+    PQ75 = pq75
+  )
+
+  # Posterior summaries computed by SAS
+  sas_sc1 <- vector("list", 1)
+  names(sas_sc1) <- names(tab_sc1)
```
danielinteractive, line 78: like if we can avoid this list code here it could be easier to read

tests/testthat/test-sas-results-part-1.R
```diff
@@ -0,0 +1,398 @@
...
+  mcmc_options <- McmcOptions(
+    burnin = 5000,
+    step = 2,
+    samples = 1000000,
+    rng_kind = "Wichmann-Hill",
+    rng_seed = 1
+  )
+
+  dose_grid_sc1 <- c(10, 20, 35, 50, 65, 80, 90, 100)
+  tab_sc1 <- vector("list")
```
danielinteractive, line 14: thanks for splitting it up @marlenesg ! nice. Now that we have that, no need for lists anymore I think here - it will be easier to read when you just e.g. use `data <- Data(...)` and similarly for the other inputs.

tests/testthat/test-sas-results-part-1.R
```diff
@@ -0,0 +1,398 @@
...
+
+  # Posterior summaries computed by SAS
+  sas_sc1 <- vector("list", 1)
+  names(sas_sc1) <- names(tab_sc1)
+
+  temp <- read.csv2(paste0(getwd(), "/testdata/sc1_sit1.csv"),
+    header = TRUE, dec = "."
+  )
+  sas_sc1[[1]] <- apply(as.matrix(temp[, -1]), 2, as.numeric)
+  rownames(sas_sc1[[1]]) <- temp[, 1]
```
danielinteractive, line 84: because now here e.g. you anyway just use the first list element always, meaning that here you don't need a list at all. you can just directly have the variable saved and work with that

danielinteractive (commented):
thanks Marlene, last comments inside. Also can you please add yourself to the authors of the package in the `DESCRIPTION` file? thanks

tests/testthat/test-sas-results-part-1.R
marlenesg: Thanks, Daniel! I changed the code accordingly

tests/testthat/test-sas-results-part-1.R
```diff
@@ -0,0 +1,367 @@
...
+    PTARGET = dose_rec_loss$probs[, "target"],
+    POVEREX = dose_rec_loss$probs[, "excessive"],
+    POVERUN = dose_rec_loss$probs[, "unacceptable"],
+    POVER = rowSums(dose_rec_loss$probs[, c("excessive", "unacceptable")]),
+    PMEAN = dose_rec_loss$probs[, "mean"],
+    PQ75 = pq75
+  )
+
+  # Posterior summaries computed by SAS
+  temp <- read.csv2(paste0(getwd(), "/testdata/sc1_sit1.csv"),
```
danielinteractive, line 72: I think for this much better to use 
```suggestion
  temp <- read.csv2(test_path("testdata/sc1_sit1.csv"),
```
and similar for the other files you read

danielinteractive: and don't forget to add yourself to package authors as mentioned above :-)

marlenesg: Thank you for the feedback, Daniel! Also just added myself to the authors :)

danielinteractive (approved):

wwojciech:
Hi @marlenesg Unit tests for SAS tests take ~30sec, see below:
```
✔ |        12 | sas-results-part-1 [32.5s]
```
Is there anything we could do to decrease this time?

@danielinteractive - FYI.

PR #497, opened 2023-03-21
487 parallel computing with extensions

### LINKED ISSUE #487: Possibility to use parallel computing with extensions to the package
It is very easy to extend the crmPack functionality by own classes and methods. That works quite well and is sufficient for analysis of study data.
When generation operation characteristics is necessary, parallel computing is necessary. However, currently the code extensions that are not part of the installed package cannot be used, i.e. any code that is used during the parallel computing needs to be part of the installed R package.
The feature request is to make user written extension usable with parallel computing without creating and installing a new package.

0liver0815:
Close #487 

I needed to add a reference in [vignettes.bib](https://github.com/Roche/crmPack/blob/ac61f1dcfa68c28af9ecb106a1fb68beb90abb58/vignettes/vignettes.bib).
I did only add to the end of the file, but for whatever reasons, some other line where changed when saving. Fortunately this concerns only a few lines, which may be ok for review.  I think that this may have to do with the line endings. Sorry for that. Please let me know if there is any possibility to avoid this in the future.


R/Design-methods.R
```diff
@@ -131,6 +131,21 @@ getResultList <- function(fun,
         varlist = ls(.GlobalEnv)
       )
 
+      # load user extensions
+      tryCatch(
+        {
+          if (exists("crmpack_extensions") == TRUE) {
```
danielinteractive, line 137: instead of doing this with a function definition as a global variable - I think it would be better to do it via a global option (https://stat.ethz.ch/R-manual/R-devel/library/base/html/options.html)

vignettes/parallel_computing_with_extensions.Rmd
```diff
@@ -0,0 +1,366 @@
...
+When trial simulations need to be performed to derive the operational characteristics of a study setup, run times may become long with single core processing. The run time depends mainly on the number of study replications and MCMC samples. In cases of long run times, utilizing parallel computing, i.e. using multiple CPU cores, can overcome these challenge and decrease run times significantly.
+
+Parallel computing is supported by `crmPack` by design, i.e. whenever the simulation call is invoked with parameters `parallel = TRUE` and `nCores =` more than one CPU, each core is initialized with the `crmPack` package and the global environment.
+
+Due to the nature of the S4 object system, user defined classes and method that were defined in the global environment cannot just made available at each core by initializing the core with the global environment. User defined classes and method needs to be executed on any core that is initialized to warrant full functionality of the S4 objects.
```
danielinteractive, line 30:
```suggestion
Due to the nature of the S4 object system, user defined classes and method that were defined in the global environment cannot just be made available at each core by initializing the core with the global environment. User defined classes and method needs to be executed on any core that is initialized to warrant full functionality of the S4 objects. 
```

vignettes/parallel_computing_with_extensions.Rmd
```diff
@@ -0,0 +1,366 @@
...
+
+Parallel computing is supported by `crmPack` by design, i.e. whenever the simulation call is invoked with parameters `parallel = TRUE` and `nCores =` more than one CPU, each core is initialized with the `crmPack` package and the global environment.
+
+Due to the nature of the S4 object system, user defined classes and method that were defined in the global environment cannot just made available at each core by initializing the core with the global environment. User defined classes and method needs to be executed on any core that is initialized to warrant full functionality of the S4 objects. 
+
+To use user written extension with parallel computing and `crmPack`, the user written code needs to be embedded in a function that must have the fixed name `crmpack_extensions`. This function is executed at any core at the time of initialization of the core, so that the user written S4 classes and methods are available along with the `crmPack` package.
```
danielinteractive, line 32: then you could adapt that here

vignettes/vignettes.bib
```diff
@@ -363,3 +363,13 @@ @Book{ggplot2
     year = {2009},
     url = {http://ggplot2.org},
   }
+
+@article{Bove2019,
```
danielinteractive, line 367:
```suggestion
@article{SabanesBove2019,
```

danielinteractive: sorry my Spanish last name :-D

danielinteractive (commented):
thanks @0liver0815 ! please see one suggestion inside. Also can you please add a test to the unit tests for this? you might need to do it in the root test folder, so not in the testthat folder. see e.g. https://github.com/insightsengineering/hermes/blob/main/tests/test_dplyr_compatibility.R for an example that does not work in testthat, but as a separate test.

R/Design-methods.R
0liver0815: I am not sure if I understand the comment. Do you mean to make sure that it is really a function and not just an object? I am not sure how to this via the option statement.

danielinteractive: Sorry that I did not give an example.
Try out this to see the mechanism:
```r
options(extensions = function(x) x + 2)
getOption("extensions")
```

0liver0815:
> thanks @0liver0815 ! please see one suggestion inside. Also can you please add a test to the unit tests for this? you might need to do it in the root test folder, so not in the testthat folder. see e.g. https://github.com/insightsengineering/hermes/blob/main/tests/test_dplyr_compatibility.R for an example that does not work in testthat, but as a separate test.

I have looked at the reference in the Hermes package, but I can not figure out how `test_dplyr_compatibility.R` is used with the `testthat` package. When I try to run the complete tests of the package, I get error messages resulting from failed tests, which stops further execution and which are not very meaningful if one is not involved into the package development.

In addition, I thought about how to write a unit test. My idea was to work with `expect_snapshot`. I tested first without any user written code extension. This works quite well using the `parallel` option in `simulate`. 
When I try to add code for a user written extension, that works when running inactively, it fails with an error message when running within `testthat`. The error message says something that an inherited method is not found. I guess that `testthat` may use a different environment, so that the user written code is not available. However, I found no solution yet.

R/Design-methods.R
danielinteractive: Thanks @0liver0815 , ok seems not super easy, but here would be my proposal:

```r
library(parallel)
# Globally, work with an option. This is safer than a global function name.
options(crmpack_extensions = function() "bla")
# Now inside getResultList(), we can check the global option and load into
# a local function.
fun <- getOption("crmpack_extensions")
cl <- parallel::makeCluster(2)
# We can export the local function to the cluster.
parallel::clusterExport(cl, "fun", environment(fun))
# And then evaluate it.
parallel::clusterEvalQ(cl, fun())
```

would that work? main point here is to avoid the need for a globally defined function of a given name, and replace it by a global option holding the function instead.

0liver0815: I think I was to quick with my assessment. After more in-depth testing, the code above do not run within the function `getResultList`.  The reasons remains unclear to me. The problem is that when `clusterEvalQ` is invoked, the function is unknown to `clusterEvalQ`, even though it is available. However, it seems that `clusterCall` solves the problem.

vignettes/parallel_computing_with_extensions.Rmd
0liver0815: I revised the whole vignette accordingly.

PR #494, opened 2023-03-17
Closes #493

wwojciech:
Closes #493

PR #498, opened 2023-03-21
Closes #496

### LINKED ISSUE #496: Cleanup: `maxDose` methods: `generic` and `IncrementsRelative`

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #496


wwojciech:
Hi @cicdguy - do you know why this automation check is hanging here? Tx!

wwojciech:
> > [skip actions] Restyle files
> 
> This is why. The bot auto-restyled your files and committed a change. The commit message skips all actions from being run recursively.
> 
> You can push an empty commit to retrigger your workflows.

Thanks! Can we somehow solve this problem permanently? It is a bit annoying of pushing some dummy commits just to overcome this....

wwojciech:
Hi @danielinteractive - it seems that `examples` were not styled in https://github.com/Roche/crmPack/pull/494. This is because styler has no option so style all the examples. So I am sorry that some of the example R files were being restyled here.

danielinteractive (commented):
Thanks Wojciech, please see inside one doubt about the tests

wwojciech:
> Thanks Wojciech, please see inside one doubt about the tests

Hi @danielinteractive - thanks, I am not able to see any comments/doubts you mentioned...

tests/testthat/test-Rules-methods.R
```diff
@@ -735,7 +735,60 @@ test_that("nextBest-NextBestMaxGainSamples returns expected values of the object
...
 })
 
-# maxDose-IncrementsNumDoseLevels ----
+# maxDose ----
+
+## IncrementsRelative ----
+
+test_that("IncrementsRelative works correctly for last dose in 1st interval", {
+  increments <- IncrementsRelative(intervals = c(0, 110), increments = c(1, 0.5))
+  data <- h_get_data(placebo = FALSE) # The last dose is 100.
```
danielinteractive, line 744: oops not sure why I did not save my comment sorry @wwojciech !
My comment was here about the `data` preparation: it would be more transparent and easy to read if you directly create the `Data()` call here incl. the last dose of 100, then also no comment needed anymore. especially for these `maxDose()` method tests. otherwise is fine to use these data helper constructors.

wwojciech: Sure, thanks @danielinteractive, I will do that. The only one small drawback is that it will be a little bit more lines of the code, but I guess it is worth to pay this price.

danielinteractive: yeah just keep the `Data()` call as minimal as possible. e.g. just one cohort at dose 100 here will be sufficient

wwojciech: yap, maybe two cohorts with two different doses - still the code it short but we can test when there are two doses so that we make sure the last one is chosen. Ok, I pushed the changes so you can have a look.

danielinteractive (approved):

wwojciech:
Hi @cicdguy, it seems that we have some problem with automation check here:
```
 "Error: processing vignette 'example.Rnw' failed with diagnostics:
pdflatex is not available
```
Can you please help on this?

PR #511, opened 2023-03-27
Closes #510

### LINKED ISSUE #510: Cleanup: `maxDose` method for `IncrementsRelativeDLT` signature.

To do:

- [ ] Revise documentation
  - [ ] Replace `##'` by `#'`
  - [ ] Add to `pkgdown.yml`
- [ ] Add unit tests
  - [ ] For classes:
  - [ ] For functions: as usual
- [ ] Refactor as needed
  - [ ] Remove `keywords` tags
  - [ ] For classes:

wwojciech:
Closes #510

### DIFF
--- a/tests/testthat/test-Rules-methods.R
+++ b/tests/testthat/test-Rules-methods.R
@@ -790,6 +790,77 @@ test_that("maxDose-IncrementsRelative throws error when Data is empty", {
   )
 })
 
+## IncrementsRelativeDLT ----
+
+test_that("maxDose-IncrementsRelativeDLT works correctly for no of DLTs in 1st interval", {
+  increments <- IncrementsRelativeDLT(dlt_intervals = c(0, 2), increments = c(1, 0.5))
+  data <- Data(
+    x = c(5, 100), y = c(0L, 0L), doseGrid = c(5, 100), ID = 1:2, cohort = 1:2
+  )
+  result <- maxDose(increments, data)
+  expect_equal(result, 200)
+
+  # 1 DLT in total.
+  data@y <- c(1L, 0L)
+  result <- maxDose(increments, data)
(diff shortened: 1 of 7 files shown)


tests/testthat/test-Rules-methods.R
```diff
@@ -790,6 +790,77 @@ test_that("maxDose-IncrementsRelative throws error when Data is empty", {
...
+  result <- maxDose(increments, data)
+  expect_equal(result, 300)
+})
+
+test_that("maxDose-IncrementsRelativeDLT throws error when no of DLTs is below the first interval", {
+  increments <- IncrementsRelativeDLT(dlt_intervals = c(2, 5), increments = c(1, 0.5))
+  data <- Data(x = c(5, 100), y = c(0L, 1L), doseGrid = c(5, 100), ID = 1:2, cohort = 1:2)
+  expect_error(
+    maxDose(increments, data),
+    "Assertion on 'dlt_happened.*increments@dlt_intervals.*failed: Must be TRUE."
```
danielinteractive, line 829: hm what are those strange error messages?... I am not sure that will be helpful. I would keep the assertion message as simple as possible. if it happens, user anyway has to debug

danielinteractive (commented):
thanks Wojciech, just one question on error message inside

tests/testthat/test-Rules-methods.R
wwojciech: Yes, unfortunately, this message comes from the `checkmate` (specifically from `assert_true()` that we use in the body of `maxDose` method(s)) and I think there is no way we can alter this. I do not like it too - but the alternative would be to use `stopifnot()` with custom error message. However, we always use `assert_*` in all the other places....

danielinteractive (approved):

PR #515, opened 2023-03-29
Closes #514

### LINKED ISSUE #514: Cleanup: `maxDose` method for `IncrementsRelativeParts` signature.

To do:

- [x] Revise documentation
  - [x] Replace `##'` by `#'`
  - [x] Add to `pkgdown.yml`
- [x] Add unit tests
  - [x] For classes:
  - [x] For functions: as usual
- [x] Refactor as needed
  - [x] Remove `keywords` tags
  - [x] For classes:

wwojciech:
Closes #514 - draft, no unit tests

### DIFF
--- a/R/Rules-methods.R
+++ b/R/Rules-methods.R
@@ -1138,70 +1138,54 @@ setMethod(
   }
 )
 
-# nolint start
+## IncrementsRelativeParts ----
 
-## --------------------------------------------------
-## The maximum allowable relative increments, with special rules for
-## part 1 and beginning of part 2, method method
-## --------------------------------------------------
+#' @describeIn maxDose determine the maximum possible next dose based on
+#'   relative increments as well as part 1 and beginning of part 2.
+#'
+#' @aliases maxDose-IncrementsRelativeParts
+#'
+#' @export
+#' @example examples/Rules-method-maxDose-IncrementsRelativeParts.R
+#'
+setMethod(
+  f = "maxDose",
+  signature = signature(
+    increments = "IncrementsRelativeParts",
+    data = "DataParts"
+  ),
+  definition = function(increments, data, ...) {
 
-##' @describeIn maxDose Determine the maximum possible next dose based on
-##' relative increments and part 1 and 2
-##' @example examples/Rules-method-maxDose-IncrementsRelativeParts.R
-setMethod("maxDose",
-  signature =
-    signature(
-      increments = "IncrementsRelativeParts",
-      data = "DataParts"
-    ),
-  def =
-    function(increments, data, ...) {
-      ## determine if there are already cohorts
-      ## belonging to part 2:
-      alreadyInPart2 <- any(data@part == 2L)
+    # If there are already cohorts belonging to part 2.
+    if (any(data@part == 2L)) {
+      callNextMethod(increments, data, ...)
+    } else {
+      # Maximum dose level so far.
+      max_dose_lev_part1 <- matchTolerance(max(data@x), data@part1Ladder)
 
-      ## if so, we just call the next higher method
-      if (alreadyInPart2) {
-        callNextMethod(increments, data, ...)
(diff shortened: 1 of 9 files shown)


wwojciech:
Hi @danielinteractive - I am submitting this PR as a draft (i.e. all ready except unit tests) since the logic of `maxDose-IncrementsRelativeParts` function is quite involved, it was not consistent with the help page for `IncrementsRelativeParts` class. So, cleaned it (also the help fro `IncrementsRelativeParts`) according to my best intuition and knowledge. Before I proceed with the unit tests, I wanted to ask you to quickly have a look at this and confirm if this is ok and I can proceed to the unit tests. Tx!

R/Rules-methods.R
```diff
@@ -1138,71 +1138,55 @@ setMethod(
...
-      increments = "IncrementsRelativeParts",
-      data = "DataParts"
-    ),
-  def =
-    function(increments, data, ...) {
-      ## determine if there are already cohorts
-      ## belonging to part 2:
-      alreadyInPart2 <- any(data@part == 2L)
+    # If there are already cohorts belonging to part 2.
+    if (any(data@part == 2L)) {
```
danielinteractive, line 1160: I would keep the variable assignment here, since that explains what it is, even without the comment

danielinteractive (commented):
looks good, thanks Wojciech

wwojciech:
Hi @danielinteractive - I think that here:
https://github.com/Roche/crmPack/blob/f041ae22d3991892257eaea30840ddbc804c73ac/R/Rules-methods.R#L1187
we need to extract the case when:
```
lastDoseLevel + increments@dlt_start >= length(data@part1Ladder)
```
I think in such a case we should call `callNextMethod(increments, data, ...)`, i.e. a usual relative increment rules apply.

What do you think?

danielinteractive:
> Hi @danielinteractive - I think that here:
> 
> https://github.com/Roche/crmPack/blob/f041ae22d3991892257eaea30840ddbc804c73ac/R/Rules-methods.R#L1187
> 
> we need to extract the case when:
> 
> ```
> 
> (lastDoseLevel + increments@dlt_start) >= length(data@part1Ladder)
> 
> ```
> 
> I think in such a case we should call `callNextMethod(increments, data, ...)`, i.e. a usual relative increment rules apply.
> 
> 
> 
> What do you think?

Thanks Wojciech, I think we cannot refer to the parent method because we still need to use the ladder slot to select the dose?

wwojciech:
Hi @danielinteractive, Could you pl. accept this PR if all looks ok? Tx!

wwojciech:
> > > Hi @danielinteractive - I think that here:
> > > https://github.com/Roche/crmPack/blob/f041ae22d3991892257eaea30840ddbc804c73ac/R/Rules-methods.R#L1187
> > > 
> > > we need to extract the case when:
> > > ```
> > > 
> > > (lastDoseLevel + increments@dlt_start) >= length(data@part1Ladder)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I think in such a case we should call `callNextMethod(increments, data, ...)`, i.e. a usual relative increment rules apply.
> > > What do you think?
> > 
> > 
> > Thanks Wojciech, I think we cannot refer to the parent method because we still need to use the ladder slot to select the dose?
> 
> yap, but how can we use it since we are already exceeding `part1Ladder` "scale", and also we are already in part2....

I created a separate issue ( #517 ) to clarify this, and will close this PR now. Thanks!

pengguanya:
@danielinteractive I just found this probably has been implemented already:
https://github.com/Roche/crmPack/blob/8a7cffa2a0263d008f321311bc615d4d64b28960/R/Rules-methods.R#L1567

Please confirm if we should close this issue.

Issue #325, opened 2022-09-09
Add no-argument constructors for all classes


danielinteractive:
Motivation: For downstream shiny and maybe other code development it could be useful to have a construction of any object without specific arguments.

To do:

- [ ] Go through all classes (extract e.g. from NAMESPACE file) and put them here in a checklist:
	- [ ] e.g. `Data`
	- [ ] etc.
- [ ] For each class, add a default no-argument constructor that prefixes the class name with `.Default`:
	- [ ] `.DefaultData()`
	- [ ] etc
- [ ] add tests (to make sure it works)
- [ ] document in the same page as the user constructor
- [ ] export those (so it can be used in add on packages / shiny apps by the user etc.)

Alternative: #260

wwojciech:
Hi @danielinteractive - if I understand this correctly, this means that some default values for the class' slots should be hardcoded in the class' user constructor? Or should we just define a default values for the slots in prototype of class, and then the user no-args constructor would be just simply `A <- function() { .A() }` (wherever possible)

danielinteractive:
@wwojciech Yes, reasonable default values should be present in this no-argument default constructor, naming convention `.DefaultClass()` where `Class` is the usual class name. If we can achieve this with prototype in the class definition, nice, if not, then need additional slot initialization.

wwojciech:
Yap, using prototype would be the best I think, but requires a bit more changes. Lets see.

PuzzledFace:
@wwojciech suggested, for example,

```
eval(parse(text=".StoppingMinCohorts()))
```

or
```
do.call(".StoppingCohortsNearDose", args = list())
```

but this currently has problems, both of validity and usefulness.  For example

```
.ModelParamsNormal()
An object of class "ModelParamsNormal"
Slot "mean":
numeric(0)

Slot "cov":
<0 x 0 matrix>

Slot "prec":
<0 x 0 matrix>
validObject(.ModelParamsNormal())
Error in eigen(x, only.values = TRUE) : 0 x 0 matrix
```
which is inconsistent with
```
validObject(.LogisticLogNormal())
```

Since `.LogisticLogNormal()@params` is a `ModelParamsNormal` object identical to the above.

danielinteractive:
Thanks @PuzzledFace , yeah I think we need a different name and definition for these constructor functions since some classes require further slot manipulation after initialization. 

Generally I would be tending to explicit definitions for these rather than automatically constructed ones.

wwojciech:
[...]

PR #538, opened 2023-04-18
Creating a factory method for crmPack classes.  

PuzzledFace:
The factory method could easily be refactored to a `.Default<class name>()` design.

References #325.


danielinteractive:
Thanks @PuzzledFace!
It would be better (according to our standard coding guidelines) to refactor this into separate `.Default<ClassName>()` functions, rather than keeping one very large function with lots of `if/else` branching.

PuzzledFace:
> Thanks @PuzzledFace! It would be better (according to our standard coding guidelines) to refactor this into separate `.Default<ClassName>()` functions, rather than keeping one very large function with lots of `if/else` branching.

OK.  I can do that.  I can see three options for the location of the constructors:

1. In `helpers.R`, as at present.  That seems inappropriate to me as the functions are no longer really helpful.
2. In the various XXXX-class.R files.  That seems to create a bit of unnecessary clutter (both in code and test code) as the new code is not central to the package overall.
3. (My preference) In a separate file, perhaps `default-constructors.R`.

The location of unit tests would follow the decision for the code itself.

Do you have a preference?  Or an alternative suggestion?

[I agree the code is dull - and long winded.  But it is useful.]

danielinteractive:
Thanks @PuzzledFace - I think it would be best in the separate class files, next to the standard user constructors. The reason is that when the class is updated by developers, then they see the constructors next to it and remember to update them. Also when a new class is added, it is easier if it is together in the same file.

PuzzledFace:
> Thanks @PuzzledFace - I think it would be best in the separate class files, next to the standard user constructors. The reason is that when the class is updated by developers, then they see the constructors next to it and remember to update them. Also when a new class is added, it is easier if it is together in the same file.

I appreciate the argument.  I can do that.

PuzzledFace:
Two minor queries as I work on the default constructors...

### Long function names
`lintr` currently reports problems when variable or function names are longer than 30 characters.  Our naming convention (prepending `.Default` to the class name) results in function names that are too long for five classes.  These are:

- `.DefaultLogisticNormalFixedMixture`
- `.DefaultLogisticLogNormalMixture`
- `.DefaultIncrementsRelativeDLTCurrent`
- `.DefaultStoppingPatientsNearDose`
- `.DefaultStoppingLowestDoseHSRBeta`

Will this cause a problem when the changes are pulled to `main`?  If so, advice, please.

### Inconsistent definition of `ref_dose`
`LogisiticNormalMixture` and `LogisiticNormalFixedMixture` require `ref_dose` to be a `positive_number`.  Other classes do not.  The restriction seems reasonable, but not mandatory.  eg In the original O'Quigley paper, many of the "doses" are not positive.  Also, when using a reference dose and `log(dose/ref_dose)`, the "doses" plugged into the model may be non-positive.

- Should we fix the inconsistency?  If so...
  - Should we tighten restrictions in other classes or loosen the requirement in `LogisiticNormalMixture` and `LogisiticNormalFixedMixture`?
  - Should I make the change as part of this pull request or should I open a second issue specific to the question?

danielinteractive:
> Two minor queries as I work on the default constructors...

Thanks @PuzzledFace !

> ### Long function names
> `lintr` currently reports problems when variable or function names are longer than 30 characters. Our naming convention (prepending `.Default` to the class name) results in function names that are too long for five classes. These are:
> 
> * `.DefaultLogisticNormalFixedMixture`
> * `.DefaultLogisticLogNormalMixture`
> * `.DefaultIncrementsRelativeDLTCurrent`
> * `.DefaultStoppingPatientsNearDose`
> * `.DefaultStoppingLowestDoseHSRBeta`
> 
> Will this cause a problem when the changes are pulled to `main`? If so, advice, please.

We can fix this by putting a  `#nolint` at the end of the first function definition line for these. That is what I would propose doing, while still keeping to the naming convention also for these classes.

> ### Inconsistent definition of `ref_dose`
> `LogisiticNormalMixture` and `LogisiticNormalFixedMixture` require `ref_dose` to be a `positive_number`. Other classes do not. The restriction seems reasonable, but not mandatory. eg In the original O'Quigley paper, many of the "doses" are not positive. Also, when using a reference dose and `log(dose/ref_dose)`, the "doses" plugged into the model may be non-positive.

Hm interesting. I just checked our `Data` class definition and indeed we don't require doses to be positive.

> * Should we fix the inconsistency?  

Yes I think that would be good

If so...
>   
>   * Should we tighten restrictions in other classes or loosen the requirement in `LogisiticNormalMixture` and `LogisiticNormalFixedMixture`?

Hm so I would loosen the requirement in these 2 classes. If this breaks any tests, we can double check.

>   * Should I make the change as part of this pull request or should I open a second issue specific to the question?

maybe good to do it separately, since it requires a few changes (not just 1 line change) and the topic is unrelated to this PR's topic

PuzzledFace:
> > Two minor queries as I work on the default constructors...
> 
> Thanks @PuzzledFace !
> 
> > ### Long function names
> > `lintr` currently reports problems when variable or function names are longer than 30 characters. Our naming convention (prepending `.Default` to the class name) results in function names that are too long for five classes. These are:
> > 
> > * `.DefaultLogisticNormalFixedMixture`
> > * `.DefaultLogisticLogNormalMixture`
> > * `.DefaultIncrementsRelativeDLTCurrent`
> > * `.DefaultStoppingPatientsNearDose`
> > * `.DefaultStoppingLowestDoseHSRBeta`
> > 
> > Will this cause a problem when the changes are pulled to `main`? If so, advice, please.
> 
> We can fix this by putting a `#nolint` at the end of the first function definition line for these. That is what I would propose doing, while still keeping to the naming convention also for these classes.
> 
> > ### Inconsistent definition of `ref_dose`
> > `LogisiticNormalMixture` and `LogisiticNormalFixedMixture` require `ref_dose` to be a `positive_number`. Other classes do not. The restriction seems reasonable, but not mandatory. eg In the original O'Quigley paper, many of the "doses" are not positive. Also, when using a reference dose and `log(dose/ref_dose)`, the "doses" plugged into the model may be non-positive.
> 
> Hm interesting. I just checked our `Data` class definition and indeed we don't require doses to be positive.
> 
> > * Should we fix the inconsistency?
> 
> Yes I think that would be good
> 
> If so...
> 
> > * Should we tighten restrictions in other classes or loosen the requirement in `LogisiticNormalMixture` and `LogisiticNormalFixedMixture`?
> 
> Hm so I would loosen the requirement in these 2 classes. If this breaks any tests, we can double check.
> 
> > * Should I make the change as part of this pull request or should I open a second issue specific to the question?
> 
> maybe good to do it separately, since it requires a few changes (not just 1 line change) and the topic is unrelated to this PR's topic

Thank you.  Will do.

PuzzledFace:
In the latest pull request, `devtools::check()` failed knitting the vignettes.  `devtools::check(vignettes=FALSE)` failed with 

```
─  checking package dependencies ...Warning in url(sprintf("%s/%s", cran, path), open = "rb") : (2.6s)
     cannot open URL 'https://rspm.apollo.roche.com/CRAN/__linux__/focal/2022-10-21/web/packages/packages.rds': HTTP status was '404 Not Found'
   Error in url(sprintf("%s/%s", cran, path), open = "rb") : 
     cannot open the connection to 'https://rspm.apollo.roche.com/CRAN/__linux__/focal/2022-10-21/web/packages/packages.rds'
   Execution halted
```

I believe this is a temporary network issue.

I've combined documentation of the `.DefaultXXXX()` constructors with the existing class documentation.  For example:
image

NEWS.md
```diff
@@ -1,4 +1,7 @@
 # Version 1.0.9000.9133
+* Added no-parameter factory functions named `.Default<class name>` to provide
```
danielinteractive, line 2:
```suggestion
* Added no-parameter constructor functions named `.Default<class name>` to provide
```

R/Model-class.R
```diff
@@ -160,6 +160,8 @@ ModelLogNormal <- function(mean, cov, ref_dose = 1) {
 #' [`LogisticNormal`] is the class for the usual logistic regression model with
 #' a bivariate normal prior on the intercept and slope.
 #'
+#' Typically, end users will not use the `.DefaultLogisticNormal()` function.
```
danielinteractive, line 163:
```suggestion
#' @note Typically, end users will not use the `.DefaultLogisticNormal()` function.
```
and I would put this directly on top of line 216 (since then again it is easier to maintain), and then similar for the other constructor documentation

R/Model-class.R
```diff
@@ -205,6 +207,17 @@ LogisticNormal <- function(mean, cov, ref_dose = 1) {
   )
 }
 
+## default constructor ----
+
+#' @rdname LogisticNormal-class
+#' @examples
+#' .DefaultLogisticNormal()
```
danielinteractive, line 214: because users won't use it, I would not put an example for these default constructors at all. Instead, you could add a small paragraph in one of the vignettes to explain this, give 1-2 examples, and highlight that it is for developers

tests/testthat/test-helpers.R
```diff
@@ -557,7 +559,7 @@ test_that("h_create_instance works correctly", {
       CohortSizeDLT(dlt_intervals = c(0, 1), cohort_size = c(1, 3))
     )
   )
-  expect_equal(h_create_instance("CohortSizeMax"), test_obj)
+  expect_equal(.DefaultCohortSizeMax(), test_obj)
```
danielinteractive, line 562: now that we have the default constructors coded next to the class definition and user constructor, it would be better for maintenance to also put the small test for each of them next to the tests for the user constructor. That way we can also avoid additional functions defined here in the test script

danielinteractive (commented):
thanks @PuzzledFace ! good direction, please see inside few comments for next steps

tests/testthat/test-helpers.R
danielinteractive: Hi @PuzzledFace would that be possible? not sure if you saw my comment therefore checking :-)

PuzzledFace: > Hi @PuzzledFace would that be possible? not sure if you saw my comment therefore checking :-)

Yes.  It's possible.  Working on it as we speak.  Currently, the tests succeed locally but fail on GitHub.  I am investigating...

danielinteractive: cool thank you! (so basically moving the tests to the specific `test-Model-class.R` etc. files and in there next to the other tests for the same class)

R/Model-class.R
PuzzledFace: If we omit the `@example` for the `.DefaultXXX()` constructors, then they won't appear in the Usage section of the online doc for each class constructor.  If that is the case, then the statement that end users typically won't  use them would be out-of-context.  Is that sensible?

I am happy to adapt one of the vignettes. Perhaps simulations or migration?

danielinteractive: Yes, I think it would be better that they don't appear in the usage section of the online docs. Because they are not intended for users, but just for developers. Then we can modify the mentioned statement or remove it.
Yes we could add that few sentences and examples then to the migration vignette, sounds good!

tests/testthat/test-helpers.R
danielinteractive: Thanks John. Let me think about it. My main challenge is that I want to minimize any kind of magic going on for tests, examples.

PuzzledFace: Aarrgghh!  The `examples` folder is listed in `.Rbuildignore`, so my testing strategy fails on GitHub because the folder isn't there to provide the expected objects for the automated testing.  So I think I will need to revert to your approach after all...

R/Model-class.R
PuzzledFace: As discussed: leave as is.  No action.

danielinteractive (approved):
thanks a lot John!!

danielinteractive:
Let me help you with resolving the conflicts now.

PR #544, opened 2023-04-24
Add nocov everywhere

cicdguy:
To see if it works.


danielinteractive:
Thanks @cicdguy , given that this is a test I won't review this (yet) :-)

PR #546, opened 2023-04-25
Add work-arounds for `covr` trace injections

### LINKED ISSUE #545: CI covr clean up
I've spent a bit of time investigating why tests are failing in `covr`. It appears that `covr` is injecting trace calls into your code (as expected), but tests which refer to the source code itself are failing due to the modified code not appearing as expected.

You can see an example of this in this [`covr` log](https://github.com/Roche/crmPack/actions/runs/4735564625/jobs/8405965825#step:12:362)

```
── Failure ('test-helpers_model.R:23:3'): h_model_dual_endpoint_sigma2W updates model components ──
`result` (`actual`) not identical to `comp` (`expected`).

body(actual$priormodel) vs body(expected$priormodel)
  `{`
  `    y ~ x + 1`
- `    if (TRUE) {`
- `        covr:::count("helpers_model.R:38:9:38:38:9:38:27518:27518")`
- `        precW ~ dgamma(precWa, precWb)`
- `    }`
+ `    precW ~ dgamma(precWa, precWb)`
  `}`
```

When `covr` runs, it injects these calls to `covr::count` throughout your code. Since this test expects that the function is un-modified, it will fail when running `covr`.

From what I can tell, the root cause is code in `R/helpers_models.R` where functions are used to define model expressions.

https://github.com/Roche/crmPack/blob/ea15a33f8353062056bfe114638185598ec4ec77/R/helpers_model.R#L35-L40

To fix this, I think the best path forward is to use `quote`s instead of `function`s to provide the model expressions. I think this would provide two benefits:

1. It will resolve some of these test failures by not relying on an unmodified function body
2. I think it's a better data structure because it provides _only_ the expression without the function call interface which doesn't seem to be used anywhere.

dgkf:
Closes #545

The approach had to change a bit. It turns out that `quote({ code })` will also be traced by `covr` so that solution didn't pan out.

Instead, the code as it is now will filter out the code injected by `covr`. It will only ever do this when covr is running, so there will be no impact on any code when run outside of `covr` just in case there are any unintended side effects while modifying the model expressions.


R/helpers_jags.R
```diff
@@ -56,23 +56,25 @@ h_jags_add_dummy <- function(object, where, dummy = 0) {
...
+
+  if (inherits(model2, "{")) {
+    model2_exprs <- as.list(model2)[-1]
+  } else {
+    model2_epxrs <- as.list(model2)
   }
+
+  body(model1) <- as.call(append(as.list(body(model1)), model2_exprs))
   model1
 }
```
dgkf, line 79: This is the primary change of the patch, accepting `model2` as a language object instead of strictly as a function.

To try to keep it backwards compatible through the refactor, it will still allow functions. I would probably encourage abandoning the function interface in its entirety in favor of passing expressions, but I'm not familiar enough with the use patterns to really know whether that would be worth the refactor.

R/helpers_model.R
```diff
@@ -85,10 +83,10 @@ h_model_dual_endpoint_rho <- function(use_fixed,
     assert_true(h_test_named_numeric(rho, permutation.of = c("a", "b")))
     comp$priormodel <- h_jags_join_models(
       comp$priormodel,
-      function() {
+      quote({
         kappa ~ dbeta(rhoa, rhob)
         rho <- 2 * kappa - 1
-      }
+      })
```
dgkf, line 89: And this is how the interface is used with quotes. Effectively it's just specifying the body of a function without having the added machinery of a function interface wrapped around it. This avoids assumptions about how the function can be used by providing a narrower interface that provides strictly the data you need.

R/helpers_jags.R
danielinteractive: Thanks Doug, it would indeed great to have backwards compatibility re: extensions of the R package by the user, so to allow them the use of functions still

R/helpers_model.R
danielinteractive: This is great! I really like it

danielinteractive:
So somehow the check does not run through? Any ideas maybe from @cicdguy ?

dgkf:
Just checking the `covr` results, and it looks like this change solved _some_, but not _all_ checks regarding the code insertions of `covr::check` calls. 

[On `main`](https://github.com/Roche/crmPack/actions/runs/4735564625/jobs/8405965825): ctrl+F for `covr:::count` gives 9 hits 
[in this PR](https://github.com/Roche/crmPack/actions/runs/4799547713/jobs/8557497083?pr=546) ctrl+F for `covr:::count` gives 6 hits

... so more investigation is needed here.

I was hoping that these would be the root cause of the test failures that error with:
```
Error parsing model file:
syntax error on line 3 near "{"
```

But it looks like either these 6 lingering issues are still the root, or there is something else.

dgkf:
Alright, realized that my little test code was the problem and `covr:::count` calls will still be injected into the `quote()` calls inside of functions.

_However_, I've found a workable solution. If we define models at the top level, they will _not_ be modified.

Just as a simple example. If we had code such as:

```r
fn <- function() {
  quote({ code })
}
```

`covr` will modify it to look like:

```r
fn <- function() {
  if (TRUE) {
    covr:::count("fn.R:2:3:2:17:3:17:4:4")
    quote({
      if (TRUE) {
        covr:::count("fn.R:2:11:2:14:11:14:4:4")
        code
      }
    })
  }
}
```

However! We can avoid this by declaring quoted code as data outside of function calls where static analysis can't differentiate them from meaningful code:

```r
priormodel <- quote({ code })
fn <- function() { priormodel }
```

I think we could even tackle this without moving to accepting quotes. That said, I think there are merits to accepting quotes that would really benefit this package but maybe that should be left for a separate interface discussion.

danielinteractive:
Cool thanks Doug!
However the last workaround does not look super pretty. I wonder if there is a potential pathway to add a new ignore statement to the `covr` syntax? I guess we might well not be the only package with this issue.

dgkf:
I ended up reworking this pretty substantially.

Solving this with a data structure change is going to add a lot of cognitive overhead to development to dance around this rather opaque use case. Instead, I added a function that will strip code of covr traces by walking the syntax tree.

In practice there were only a few places within the package where we need to pass function code through this filter: in `h_jags_join_models` and `h_jags_write_models`.

As well, I had to run a few functions generated in the test suite through this.

> I wonder if there is a potential pathway to add a new ignore statement to the covr syntax? 

I think we're on the same page here, but please take a look at the solution I landed on and let me know if you think it suits you.

R/helpers_jags.R
```diff
@@ -68,6 +68,11 @@ h_jags_join_models <- function(model1, model2) {
   assert_class(body(model1), "{")
   assert_class(body(model2), "{")
 
+  # workaround needed to avoid bugs related to covr-injected code
+  if (h_covr_active()) {
+    body(model2) <- h_covr_detrace(body(model2))
+  }
```
dgkf, line 74: This is what it looks like when used within `h_jags_join_models` to avoid injection of covr traces while building our model code.

tests/testthat/test-Model-methods.R
```diff
@@ -7,6 +7,7 @@ test_that("doseFunction-GeneralModel returns correct dose function", {
   samples <- Samples(list(alpha0 = 1, alpha1 = 2), options = McmcOptions(samples = 1))
 
   dose_fun <- doseFunction(model, alpha0 = 1, alpha1 = 2)
+  dose_fun <- h_covr_detrace(dose_fun)
```
dgkf, line 10: A few of the tests, primarily in `tests/testthat/test-Model-methods.R` needed to be run through `h_covr_detrace` to strip out any covr injections while testing the internals.

danielinteractive:
Thanks a lot @dgkf ! Yeah that looks much better in the sense that it will be much easier to develop the package like this.
Maybe still worth double checking in the `covr` code whether it could be a small change to ignore those functions where we say e.g. `#notrace` or something similar? (in the spirit of improving the world for everyone 😄 )

dgkf:
> Maybe still worth double checking in the covr code whether it could be a small change to ignore those functions where we say e.g. #notrace or something similar?

I looked pretty exhaustively through the `covr` source to see if there was anything that allowed us to introduce exclusions to trace injection, but there isn't much we can do to avoid it. There is one exception made for [`rlang`'s `{{`](https://github.com/r-lib/covr/blob/main/R/trace_calls.R#L51-L52) syntax... however `testthat` then uses `rlang` to evaluate test expressions and tries to interpolate it in the tests which adds even another layer to the mental burden of maintaining these tests for `covr`.

After looking into it, I feel like `covr`'s approach is right here. The problems we're seeing raise more concern to me over the choice of using function bodies to write JAGS code. I was thinking a quote would be the right path, but that was before I fully understood how they're used - effectively being used to write out a JAGS script. I know it doesn't feel as fancy, but I think a string would appropriately communicate the syntactic ambiguity of passing the code off to JAGS.

danielinteractive:
> I looked pretty exhaustively through the `covr` source to see if there was anything that allowed us to introduce exclusions to trace injection, but there isn't much we can do to avoid it. There is one exception made for [`rlang`'s `{{`](https://github.com/r-lib/covr/blob/main/R/trace_calls.R?rgh-link-date=2023-04-26T19%3A51%3A32Z#L51-L52) syntax... however `testthat` then uses `rlang` to evaluate test expressions and tries to interpolate it in the tests which adds even another layer to the mental burden of maintaining these tests for `covr`.

So we could not easily piggy-back on that conditioning for `rlang`'s case? E.g. introduce a new `#nosomething` syntax?

> After looking into it, I feel like `covr`'s approach is right here. The problems we're seeing raise more concern to me over the choice of using function bodies to write JAGS code. I was thinking a quote would be the right path, but that was before I fully understood how they're used - effectively being used to write out a JAGS script. I know it doesn't feel as fancy, but I think a string would appropriately communicate the syntactic ambiguity of passing the code off to JAGS.

Yeah I don't remember exactly why I started using function bodies initially to pass code to JAGS honestly. It must have been a previous interface from the R/JAGS passing. But actually looking what we have right now in https://github.com/Roche/crmPack/blob/1606f13a76d34ee1d8caf12be053b9fe0caf1fa4/R/helpers_jags.R#L184, we do convert the bodies ourselves to strings before writing them to a file. The only complication is that we are handling scientific notation conversion with https://github.com/Roche/crmPack/blob/1606f13a76d34ee1d8caf12be053b9fe0caf1fa4/R/helpers.R#L757. If we had the model syntax already in a string, not sure how we could perform this conversion anymore?

dgkf:
> So we could not easily piggy-back on that conditioning for rlang's case?

The `rlang` case uses the syntax tree to find instances of `{{`, so it has no information about comments that would be needed to add handling for something like `# notrace`. It would be possible to add, but would require entirely new machinery for `covr` to have this info available when adding tracing calls to the package. My feeling is that the need for it is rare enough that it doesn't warrant the feature enhancement.

Another option would be to contribute an option to disable tracing of `quote`, `bquote` and `expression` calls. But ultimately these are just heuristics, since any code in R can be non-standard-evaluated. Ultimately, there really isn't a good line to draw that works in all cases. 

> The only complication is that we are handling scientific notation conversion with 

Maybe I'm misunderstanding. If the code is provided as a string, why not just use the JAGS style for numeric notation from the start? Either way, I think a regex could realistically handle this scenario if it was still needed.

I think if I were to rework this behavior, I'd use JAGS strings internally within the package to avoid ambiguity and NSE issues, then provide an interface for users to specify JAGS code by providing an R expression. Users could fall back on providing JAGS code as a string when the JAGS code isn't syntactically valid R code such as when truncating distributions:

```r
# JAGS
#   lambda_1 ~ dnorm(0, s_lambda^-2)T(0,)

function() { lambda_1 ~ dnorm(0, s_lambda^-2)T(0,) }
# Error: unexpected symbol in "function() { lambda_1 ~ dnorm(0, s_lambda^-2)T"
```

danielinteractive:
Thanks @dgkf for your thoughts here, really much appreciated!

So all in all, seems we cannot or don't need to change the behavior of `covr`, on the other hand we can keep most of the code in `crmPack` as is but strip away the `covr` insertions if they are there, or avoid insertions by using strings internally.

At this stage, where we try to get soon to a validated solution I would tend to take the pragmatic route and strip away the `covr` insertions when needed, so that would be this PR. That minimizes the package changes and still gives us correct coverage calculations.

dgkf:
> At this stage, where we try to get soon to a validated solution I would tend to take the pragmatic route and strip away the covr insertions when needed, so that would be this PR. That minimizes the package changes and still gives us correct coverage calculations.

Yep - agree with you here. This is what's implemented in this PR. I think these changes highlighted some good food for thought for `crmPack`'s future, but would not want to tackle it without being really thoughtful about managing the user experience of possibly breaking changes.

danielinteractive:
Btw @dgkf don't forget to add your name to the authors list please!

R/helpers_covr.R
```diff
@@ -0,0 +1,54 @@
+#' Determine whether `covr` is currently running
+#'
```
danielinteractive, line 2: Maybe good to add `@keywords internal` to the internal functions?

R/helpers_covr.R
```diff
@@ -0,0 +1,54 @@
...
+h_covr_active <- function() {
+  identical(Sys.getenv("R_COVR"), "true")
+}
+
+#' Remove `covr` traces from an expression
+#'
+#' '`covr` interferes with expressions (often function bodies) as a primary
```
danielinteractive, line 9:
```suggestion
#' `covr` interferes with expressions (often function bodies) as a primary
```

R/helpers_covr.R
```diff
@@ -0,0 +1,54 @@
...
+#' interface of this package. To work with `covr`, we need to strip out the
+#' code that `covr` injects into the package before using these expressions
+#' for internal logic such as writing model files.
+#'
+#' This method is non-exhaustive, covering only a subset of `covr`'s tracing
+#' behaviors necessary for this package.
+#'
+#' @param expr (`language`)\cr an R expression or call to strip of `covr` trace
```
danielinteractive, line 17: that reminds me we still should transform this package to using `roxytypes` and `roxylint`

R/helpers_covr.R
```diff
@@ -0,0 +1,54 @@
...
+
+  detrace_call <- function(x) {
+    x[[3]][[3]]
+  }
+
+  detrace <- function(x) {
+    if (is_covr_trace(x)) x <- detrace_call(x)
+    if (is.call(x)) x[-1] <- lapply(x[-1], detrace)
+    x
+  }
```
danielinteractive, line 51: move those functions outside? so we can better test them?

R/helpers_covr.R
```diff
@@ -0,0 +1,54 @@
...
+    return(expr)
+  }
+
+  if (is.function(expr)) {
+    body(expr) <- h_covr_detrace(body(expr))
+    return(expr)
+  }
+
+  is_covr_trace <- function(x) {
+    # matches `if (TRUE) { covr:::count(<trace>); <expr> }` (see covr:::trace_calls)
```
danielinteractive, line 33:
```suggestion
    # Matches `if (TRUE) { covr:::count(<trace>); <expr> }` (see covr:::trace_calls).
```
Sentence style for comments

R/helpers_jags.R
```diff
@@ -68,6 +68,11 @@ h_jags_join_models <- function(model1, model2) {
   assert_class(body(model1), "{")
   assert_class(body(model2), "{")
 
+  # workaround needed to avoid bugs related to covr-injected code
```
danielinteractive, line 71:
```suggestion
  # This workaround is needed to avoid bugs related to covr-injected code.
```

R/helpers_jags.R
```diff
@@ -185,6 +191,11 @@ h_jags_write_model <- function(model, file = NULL, digits = 5) {
   assert_function(model)
   assert_count(digits)
 
+  # workaround needed to avoid bugs related to covr-injected code
```
danielinteractive, line 194:
```suggestion
  # This workaround is needed to avoid bugs related to covr-injected code.
```

danielinteractive (commented):
Thanks a lot Doug, just added a few review comments. If you could add some small tests for the covr helpers too that would be awesome.

R/helpers_covr.R
dgkf-roche: I moved all but `detrace` outside. My thinking for keeping `detrace` here is that it's only purpose is to avoid repeatedly testing for whether covr is active while recursing. This could live in a separate function, but it's small enough as-is that it doesn't decouple much functionality.

Otherwise, everything else now has its own helper function.

R/helpers_covr.R
dgkf: Going to mark this one as resolved, but this could certainly be a separate cleanup task. I'm curious to hear how they're working out for `savvyr` or any other repos you've adopted them for.

PR #567, opened 2023-05-15
Jk improve test coverage

PuzzledFace:
Test coverage for Rules-methods.R is now >87%.  Tests are still needed for
* `stopTrial` for
    * `StoppingTargetBiomarker`
      * With column names containing “Emax”
      * `is_relative == FALSE`
    * `CIMaxGainSamples`
    * `CIMaxGain`

The `stopTrial` function for `CIMaxGainSamples` and `CIMaxGain` is complex.  I'm not sufficiently familiar with the models to do anything other than copy and paste code to reproduce results from the current implementation.

The test for `stopTrial` applied to  `StoppingTargetBiomarker` is similarly crude and simply tests that the results of the current example are reproduced.  Note that the example code produces a `stopTrial` result of `FALSE` for all doses in the example dose grid.


PuzzledFace:
Test failures appear to be all related to comparison with SAS outputs.

PR #574, opened 2023-05-23
555 clean da design

### LINKED ISSUE #555: Clean up and test `DADesign`
See https://github.com/Roche/crmPack/blob/main/R/Design-class.R#L480 and corresponding model `DALogisticLogNormal`

To do:
- [x] Clean up similarly as the other classes above in the file
    - [x] Add for simpler navigation in RStudio the lines with `# ClassName ----` and `## class ----` and `## constructor ----`
	- [x] includes changing to `snake_case` where it helps: here change `Effmodel` to `eff_model`
	- [x] documentation cleanup:
    	- [x] replace double `##` with `#`
    	- [x] remove `@keywords class` line
    	- [x] replace `\code{\linkS4class{ClassName}}` with ``[`ClassName`]``
    	- [x] Use ``@slot model (`type or class`)\cr`` for documenting the type or class of the slots
    	- [x] Use simple title `ClassName` followed by the lifecycle badge, followed by the description of the class
    	- [x] Add the `@aliases` line for the class definition
	- [x] Use the argument names for `setClass` and replace `representation()` with simple `c()`
	- [x] For `contains` just use simple string instead of list with a single string
	- [x] remove the `validObject` line
	- [x] user constructor:
		- [x] Move the example here, and for consistency rename the file to start with upper case letter, e.g. `Design` instead of `design`
		- [x] Use `@rdname ClassName-class` to put the constructor documentation in the same place as the class definition
		- [x] Use ``@param model (`type or class`)\cr see slot definition.`` for documenting the arguments
		- [x] Remove `@keywords methods` line
		- [x] Remove `@return` line for now 
		- [x] Use `@inheritDotParams RuleDesign`
- [x] Clean up the example code file
- [x] Add unit tests to https://github.com/Roche/crmPack/blob/main/tests/testthat/test-Design-class.R
    - [x] compare other tests in that file and use same approach

pengguanya:
Closes #555


R/Design-class.R
```diff
@@ -409,110 +409,118 @@ DualResponsesSamplesDesign <- function(eff_model,
...
+## class ----
+
+#' `DualResponsesDesign`
+#'
+#' @description `r lifecycle::badge("stable")`
+#'
+#' This is a class of design based on DLE responses using the [`LogisticIndepBeta`] model
+#' model and efficacy responses using [`ModelEff`]  model class
+#' without DLE and efficacy samples. It contain all slots in
+#' [`RuleDesign`] and [`TDDesign`] class object
```
danielinteractive, line 423:
```suggestion
#' This is a class of designs based on DLE responses using the [`LogisticIndepBeta`] model
#' and efficacy responses using the [`ModelEff`]  model, without DLE and efficacy samples. 
#' It contains all slots from the [`RuleDesign`] and [`TDDesign`] classes, as it inherits from
#' [`TDDesign`].
```

R/Design-class.R
```diff
@@ -409,110 +409,118 @@ DualResponsesSamplesDesign <- function(eff_model,
...
+#'
+#' @slot data the data set of [`DataDual`] class object
+#' @slot eff_model the pseudo efficacy model to be used, an object class of
+#' [`ModelEff`]
```
danielinteractive, line 427:
```suggestion
#' @slot data (`DataDual`)\cr the data.
#' @slot eff_model (`ModelEff`)\cr the pseudo efficacy model.
```

R/Design-class.R
```diff
@@ -409,110 +409,118 @@ DualResponsesSamplesDesign <- function(eff_model,
...
+#'
+#' @export
+#'
 .DualResponsesDesign <-
   setClass(
     Class = "DualResponsesDesign",
-    representation(
-      Effmodel = "ModelEff",
+    representation = list(
```
danielinteractive, line 434:
```suggestion
    slots = c(
```

R/Design-class.R
```diff
@@ -409,110 +409,118 @@ DualResponsesSamplesDesign <- function(eff_model,
...
-##' @param \dots additional arguments for \code{\link{TDDesign}}
-##' @return the \code{\linkS4class{DualResponsesDesign}} class object
-##'
-##' @export
-##' @keywords methods
-DualResponsesDesign <- function(Effmodel,
+
+## constructor ----
+
+#' @rdname DualResponsesDesign-class
```
danielinteractive, line 449:
```suggestion
#' @rdname DualResponsesDesign-class
#'
```

R/Design-class.R
```diff
@@ -409,110 +409,118 @@ DualResponsesSamplesDesign <- function(eff_model,
...
+#' @param data please refer to [`DualResponsesDesign`] class object
+#' @param eff_model please refer to [`DualResponsesDesign`] class object
```
danielinteractive, line 451:
```suggestion
#' @param data (`DataDual`)\cr see slot definition.
#' @param eff_model (`ModelEff`)\cr see slot definition.
```

R/Design-class.R
```diff
@@ -409,110 +409,118 @@ DualResponsesSamplesDesign <- function(eff_model,
...
+#' @param \dots additional arguments for \code{\link{TDDesign}}
```
danielinteractive, line 452:
```suggestion
```

R/Design-class.R
```diff
@@ -409,110 +409,118 @@ DualResponsesSamplesDesign <- function(eff_model,
...
+#' @inheritDotParams RuleDesign
+#'
+#' @example examples/design-class-DualResponsesDesign.R
```
danielinteractive, line 455:
```suggestion
#' @example examples/Design-class-DualResponsesDesign.R
```
and please rename this file accordingly

R/Design-class.R
```diff
@@ -409,110 +409,118 @@ DualResponsesSamplesDesign <- function(eff_model,
...
-##' time-to-DLT data
-##' @slot data what is the dose grid, any previous data, etc., contained
-##' in an object of class \code{\linkS4class{DataDA}}
-##' @slot safetyWindow still to be documented.
-##'
-##' @example examples/design-class-DADesign.R
-##' @export
-##' @keywords classes
+
+# DADesign ----
```
danielinteractive, line 469: Oh. Actually that was your task, not the class above! Note that Dimitris has taken on the `DualResponsesDesign` cleanup task. Please can you revert the changes above and only clean the `DADesign` part?

R/Design-class.R
```diff
@@ -409,110 +409,118 @@ DualResponsesSamplesDesign <- function(eff_model,
...
+
+## class ----
+
+#' `DADesign`
+#'
+#' @description `r lifecycle::badge("stable")`
+#'
+#' This class has special requirements for the \code{model} and \code{data}
```
danielinteractive, line 477:
```suggestion
#' This class has special requirements for the `model` and `data`
```
we want to use the markdown syntax

R/Design-class.R
```diff
@@ -409,110 +409,118 @@ DualResponsesSamplesDesign <- function(eff_model,
...
+#' slots in comparison to the parent class [`Design`]:
+#'
+#' @slot model (`DADesign`), an object of or inheriting from class
+#' [`GeneralModel`], see in particular
+#' [`DALogisticLogNormal`] and
+#' [`TITELogisticLogNormal`] which make use of the
+#' time-to-DLT data
+#' @slot data what is the dose grid, any previous data, etc., contained
+#' in an object of class [`DataDA`]
+#' @slot safetyWindow still to be documented.
```
danielinteractive, line 487:
```suggestion
#' @slot model (`GeneralModel`)\cr the model to use, see in particular [`DALogisticLogNormal`] and
#' [`TITELogisticLogNormal`] which make use of the time-to-DLT data.
#' @slot data (`DataDA`)\cr what is the dose grid, any previous data, etc.
#' @slot safetyWindow (`SafetyWindow`)\cr the safety window to apply between cohorts.
```

R/Design-class.R
```diff
@@ -409,110 +409,118 @@ DualResponsesSamplesDesign <- function(eff_model,
...
+#'
+#' @aliases DADesign
+#' @export
+#'
 .DADesign <-
   setClass(
     Class = "DADesign",
-    representation(
+    representation = list(
```
danielinteractive, line 495:
```suggestion
    slots = c(
```

R/Design-class.R
```diff
@@ -409,110 +409,118 @@ DualResponsesSamplesDesign <- function(eff_model,
...
+
+## constructor ----
+
+#' @rdname DADesign-class
+#'
+#' @param model (`DALogisticLogNormal`)\cr see slot definition.
+#' @param data see [`DADesign`]
+#' @param safetyWindow see [`DADesign`]
+#' @param \dots additional arguments for \code{\link{Design}}
+#' @inheritDotParams RuleDesign
```
danielinteractive, line 518:
```suggestion
#' @param model (`GeneralModel`)\cr see slot definition.
#' @param data (`DataDA`)\cr see slot definition.
#' @param safetyWindow (`SafetyWindow`)\cr see slot definition.
#' @inheritDotParams Design
```

examples/Design-class-DADesign.R
```diff
@@ -1,10 +1,11 @@
 # nolint start
```
danielinteractive, line 1: please remove the `nolint` start and end because we now want this file to be checked by the linter

examples/Design-class-DADesign.R
```diff
@@ -1,10 +1,11 @@
...
 
 # Define the dose-grid and PEM parameters
```
danielinteractive, line 3: I would remove all comments

tests/testthat/test-Design-class.R
```diff
@@ -317,3 +317,76 @@ test_that("DualResponsesSamplesDesign user constructor arguments names are as ex
     ordered = TRUE
   )
 })
+
+# DADesign ----
+
+test_that(".DADesign works as expected", {
+  # Create an instance of .DADesign
```
danielinteractive, line 324: looks nice, good work! Can you just remove the comments? (see above where we also don't have them)
(Generally, we avoid comments where not absolutely necessary, as we intend to have readable code instead.)

danielinteractive (commented):
Thanks @pengguanya , nice work, please see inside, in particular please leave the `DualResponsesDesign` stuff as it was since we have another task ongoing for that. Please let me know if any questions then we can have a quick chat any time :-)

R/Design-class.R
pengguanya: As you suggested, I will keep all DualResponsesDesign as it was in main branch. So this change will not be implemented.

R/Design-class.R
pengguanya: Yes, just done that. DualResponsesDesign part (line 412 to 463) is now exactly as in main branch

R/Design-class.R
pengguanya: @danielinteractive 
Few questions:
- Am I right that the document for the constructor function of `DADesign` is not meant to be accessible or read by the user, because what they really care is the class?
- If so, why do we add all these details in the header? Is that for the developer? 
- How to read document specifically for the constructor function from rstudio? When I do `?DADesign` in rstudio, I only see the document for the class (composed by the header of .DADesign class)
- What is the difference between ( ) and [ ]?

danielinteractive: The user constructor (called `DADesign()`) is to be used by the user directly, therefore we export it and document it in the same place. We want that this is in the same place, so when calling `?DADesign` it should all be there - the class documentation and the user constructor.

For the markdown, `()` is just parentheses, but `[]` means a link to the documentation page. We use those differently as you can see from the other roxygen2 chunks.

R/Design-class.R
```diff
@@ -524,15 +523,3 @@ DADesign <- function(model,
...
-  data = DataDA(doseGrid = 1:2),
-  safetyWindow = .SafetyWindowConst(),
-  nextBest = .NextBestNCRM(),
-  startingDose = 1,
-  cohortSize = CohortSizeConst(3),
-  stopping = StoppingMinCohorts(10),
-  increments = IncrementsDoseLevels(2)
-))
-
-# nolint end
```
danielinteractive, line 538: maybe you need to move the `nolint end` line just above `DADesign` now since that part up there is not yet cleaned

tests/testthat/test-Design-class.R
```diff
@@ -317,3 +317,67 @@ test_that("DualResponsesSamplesDesign user constructor arguments names are as ex
...
```
danielinteractive, line 324: in this test I still see the unnecessary comments

tests/testthat/test-Design-class.R
```diff
@@ -317,3 +317,67 @@ test_that("DualResponsesSamplesDesign user constructor arguments names are as ex
...
+      data = data,
+      nextBest = next_best,
+      safetyWindow = safety_window,
+      cohortSize = cohort_size,
+      startingDose = starting_dose,
+      stopping = stopping,
+      increments = increments
+    )
+  )
+
```
danielinteractive, line 361:
```suggestion
```

tests/testthat/test-Design-class.R
```diff
@@ -317,3 +317,67 @@ test_that("DualResponsesSamplesDesign user constructor arguments names are as ex
...
+  expect_true(identical(result@data, data))
+  expect_true(identical(result@safetyWindow, safety_window))
+  expect_true(identical(result@nextBest, next_best))
+  expect_true(identical(result@cohortSize, cohort_size))
+  expect_true(identical(result@startingDose, starting_dose))
+  expect_true(identical(result@stopping, stopping))
+  expect_true(identical(result@increments, increments))
+})
+
+test_that("Design user constructor arguments names are as expected", {
```
danielinteractive, line 377: please test the `DADesign` arguments, not `Design`

danielinteractive (commented):
thanks @pengguanya , much better now, see last comments inside

R/update.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
```
danielinteractive, line 1: ah nice - but please move this to another (internal repo) folder, e.g. `devel` could be suitable

PR #575, opened 2023-05-24
Jk simulation vignette

Puzzled-Face:
Introducing a new vignette that describes basic sanity checking on a  candidate design (as discussed at the team meeting on 23-May-2023.  This is intended to be a precursor to the full simulation vignette.

I have made consequential changes to README and WORDLIST.


vignettes/trial_sanity_checks.Rmd
```diff
@@ -0,0 +1,338 @@
...
+
+maxDose(revised_increment_rule, no_tox_below_50_data)
+```
+
+Whilst the increments rule allows escalation to 83.5 mg, the toxicity estimates provided by the model do not: the current estimate of toxicity at 80 mg is `r recommended_dose$probs[10, 3]`, well above the limit of 0.25 defined in the dose recommendation rule (`next_best`).
+
+Again, we have several options.  We can introduce intermediate doses, thus slowing escalation and allowing the prior model more time to adapt (but probably requiring a larger trial), or we can modify the prior so that it allows faster, but still reasonable, escalation.  Let's try creating a minimally informative prior that is consistent with our first attempt.
+
+```{r, error=TRUE}
+# Fitting the min-inf_model is slow.
```
danielinteractive, line 223: it would be good for this chunk to avoid commented code and instead work with an if/else condition - if the save file is there, skip the code execution and load it, otherwise execute the code and save to the file.

danielinteractive (commented):
very nice, thanks John!

danielinteractive (approved):
Thanks @Puzzled-Face , I think you also need to update https://github.com/Roche/crmPack/blob/main/_pkgdown.yaml#L14. Otherwise good to go from my side

Puzzled-Face:
> Thanks @Puzzled-Face , I think you also need to update https://github.com/Roche/crmPack/blob/main/_pkgdown.yaml#L14. Otherwise good to go from my side

Done.  Good catch.  One I overlooked.  Thanks.

PR #584, opened 2023-06-02
Add CamelCase snake_case motivation

### LINKED ISSUE #401: Snake case and Camel case: add some background explanation and logic
The goal after the cleanup is that classes are using Camel case and parameters and slots are using Snake case. We should add some explanation and the logic to the documentation so that new and old users understand the logic and old users are aware that their code may need change in parameter names to run with the new version.

To do:
- [ ] Add paragraph on this to the relevant part in https://github.com/Roche/crmPack/blob/main/vignettes/migration_from_the_old_crmPack.Rmd

0liver0815:
Closes #401

### DIFF
--- a/vignettes/migration_from_the_old_crmPack.Rmd
+++ b/vignettes/migration_from_the_old_crmPack.Rmd
@@ -22,13 +22,13 @@ The following vignette describes the changes that were introduced to the `crmPac
 package as a result of the package's refactoring. One row in below tables
 represents a single-type, consistent change.
 
-# Class and slot changes
+## Class and slot changes
 
-## Naming convention motivation
+### Naming convention motivation
 
 To be close to common R style guidelines [@advancedR] and use consistent naming conventions within the `crmPack` package, `CamelCase` notation is used for `class` names, `method` names and constructor function names, and `snake_case` notation is used for `slot` names throughout the package.
 
-## New classes
+### New classes
 
 ```{r new_classes, echo=FALSE}
 new_classes <- tribble(
@@ -79,7 +79,7 @@ kbl(cn, longtable = TRUE, booktabs = TRUE, col.names = NULL, escape = FALSE) %>%
   kable_styling(latex_options = c("repeat_header"), font_size = 11)
 ```
 
-## Renamed classes
+### Renamed classes
 
 ```{r renamed_classes, echo=FALSE}
 renamed_classes <- tribble(
@@ -100,7 +100,7 @@ kbl(cr, longtable = TRUE, booktabs = TRUE, col.names = NULL, escape = FALSE) %>%
   kable_styling(latex_options = c("repeat_header"), font_size = 11)
 ```
 
-## Renamed slots
+### Renamed slots
 
 ```{r renamed_slots, echo=FALSE}
 #   "helper", "positive_number", "", "", FALSE, "to handle strictly positive valued slots
@@ -209,12 +209,12 @@ Strikeout indicates that the class/slot was removed.
 
 
 
-# Moved `dose` and `prob` Functions from Slots to Methods
+## Moved `dose` and `prob` Functions from Slots to Methods
 
 Moved `dose` and `prob` functions from model class slots to model class methods.
 Example of usage: `dose`/`prob` function as a true dose-DLT/DLT-dose relationship.
 
-## Generate data, define a model and get samples
+### Generate data, define a model and get samples
 
 ```{r data_model_samples}
 library(crmPack)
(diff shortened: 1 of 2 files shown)


vignettes/migration_from_the_old_crmPack.Rmd
```diff
@@ -46,10 +51,10 @@ class_changes <- tribble(
   "DALogisticLogNormal", "", "C_par", "c_par", FALSE, "",
   "DALogisticLogNormal", "", "conditionalPEM", "cond_pem", FALSE, "",
   "TITELogisticLogNormal", "", "weightMethod", "weight_method", FALSE, "",
-  "OneParLogNormalPrior", "", "skeletonFun", "skel_fun", FALSE, "",
+  "OneParLogNormalPrior", "", "", "skel_fun", FALSE, "",
```
danielinteractive, line 54: are these other changes here intended?

danielinteractive (commented):
thanks @0liver0815 , looks good, just double checking the table changes

0liver0815:
When writing the explanation a came across the following:

- Unfortunately we are not consisting with the naming. A new `class` `positive_number` is introduced, which do not follow the `CamelCase` convention for `class`.
- We do not follow strictly follow `UpperCamelCase`, but also not `lowerCamelCase`, so that `method` names like `biomarker` do not fit. However, this is probably neglectable.
- The content of the vignette do not only cover changes from 1.0.3 to the new version, but also new `classes` as `OneParLogNormalPrior` or `TITELogisticLogNormal`. Probably we need to split the vignette or at least have separate sections for the migration and new classes.

@danielinteractive please let me know if you agree. I can create an issue and work on separating new `classes` and `methods` from changed ones.

vignettes/migration_from_the_old_crmPack.Rmd
0liver0815: Yes, please see my comment below. `OneParLogNormalPrior` is new, e.g. not part of 1.0.3. However we changed the name during development, but that is nothing we need to track (in my opinion).

danielinteractive:
> When writing the explanation a came across the following:
> 
> * Unfortunately we are not consisting with the naming. A new `class` `positive_number` is introduced, which do not follow the `CamelCase` convention for `class`.

Hm right. Maybe add an "Exceptions" paragraph or so and explain that here we want the name close to "numeric" etc.

> * We do not follow strictly follow `UpperCamelCase`, but also not `lowerCamelCase`, so that `method` names like `biomarker` do not fit. However, this is probably neglectable.

Yeah can neglect or add to exceptions. 

> * The content of the vignette do not only cover changes from 1.0.3 to the new version, but also new `classes` as `OneParLogNormalPrior` or `TITELogisticLogNormal`. Probably we need to split the vignette or at least have separate sections for the migration and new classes.

OK good to split into separate tables, would keep in the same vignette though
> 
> @danielinteractive please let me know if you agree. I can create an issue and work on separating new `classes` and `methods` from changed ones.

maybe can do here, since this PR cleans up the migration vignette a bit more anyway

0liver0815:
Will do. I will let you when a review makes sense.

vignettes/migration_from_the_old_crmPack.Rmd
danielinteractive: correct, agree

0liver0815:
@danielinteractive I think you can review now. I have rewritten the first section from the user perspective, i.e. the user needs to know about new classes, renamed classes and changed slots, but not necessarily about changes in methods. If I am not mistaken, a user may not use methods directly so, changes in parameter names are not important. I moved the section to the end of the vignette. In addition generic classes will not directly be used by the user and need to be be mentioned for migration. I moved that table also to the end and removed any class that were not part of v1.03 or mentioned already at the beginning of the vignette.

vignettes/migration_from_the_old_crmPack.Rmd
```diff
@@ -21,7 +22,283 @@ The following vignette describes the changes that were introduced to the `crmPac
...
+
+class_name <- "LogisticNormal"
+eval(parse(text = paste0(".Default", class_name, "()")))
+```
+
+# Handling of `NA` or placebo returned as next dose
+
+For consistent handling how the study is stopped and to facilitate analysis of stop reasons in the operation characteristics, the handling of `NA` and placebo returned by `nextBest` methods is changed. In the previous version of `crmPack` stopping for placebo or `NA` returned by a `nextBest` method was handled automatically in the generic `Stopping` method. This is now moved into a new stopping rule `StoppingMissingDose`. As a consequence, the stopping rule `StoppingMissingDose` must be specified for those `nextBest` methods that can return NA, or when placebo is used. Otherwise the simulation may run into an error if the study is not stopped when  `NA` is returned as the next dose. `nextBest` methods that can return `NA` are `NextBestNCRM`, `NextBestNCRMLoss` and `NextBestDualEndpoint`.
+
+# Evaluation of stopping rules at a specific dose
```
danielinteractive, line 296: Can you please double check the levels of the section headers? e.g. this I don't think it should be top level? (but `##` or `###` even)

danielinteractive (commented):
Thanks @0liver0815 for the updates! Sounds great. Just one question on the section hierarchy

danielinteractive (approved):

PR #588, opened 2023-06-07
583 stopping rule na handling

### LINKED ISSUE #583: Factor out `is.na(dose)` logic from `stopTrial` generic function into separate Stopping rule
To do:

- [ ] Add unit tests for all `Stopping` classes to check that they can deal with `NA` dose values
- [x] Simplify the  `stopTrial` generic function
	- [x] see https://github.com/Roche/crmPack/blob/main/R/Rules-methods.R#L1796
	- [x] cut the block from line 1799 to line 1809 so that we have a standard generic
- [x] Add a new `Stopping` class `StoppingMissingDose`
	- [x] this should stop when the dose produced by the `NextBest` rule is `NA` 
	- [x] docs
	- [x] test
- [ ] Add references to the class from the relevant `NextBest` classes that can produce `NA`
- [ ] Add a paragraph to the migration vignette explaining this change and highlighting for which `NextBest` classes this rule needs to be added

0liver0815:
Close #583

This PR is nor final yet, but it would be nice to get input on the following:
+ Is the new class `StoppingMissingDose` created as expected?
+ Is the NA handling for Stopping rules `StoppingPatientsNearDose`, `StoppingMTDdistribution` and `StoppingHighestDose` as expected?
+ What is the strategy for NA handling for `StoppingTargetProb` (may need more change)?

In addition this PR contains many changes in the test-Rules-method.R file:
+ Add NA tests for most of the stopping rules (missing ones see below). 
+ Add markup headers for easy navigation.
+ Order stopping class in order of Rules-class.R to be able to check for completeness.

Todo:
+ Add NA handling for `StoppingTargetProb`, `StoppingTargetBiomarker`, `StoppingTDCIRatio` and `StoppingMaxGainCIRatio`.
+ For some stopping rules checks are completely missing or additional checks should be added, e.g. `StoppingHighestDose`, `StoppingMaxGainCIRatio`


danielinteractive (commented):
Thanks @0liver0815 great start, thanks for taking this on!

- Yes, the new class `StoppingMissingDose` is created as expected
- Yes, NA handling for Stopping rules `StoppingPatientsNearDose`, `StoppingMTDdistribution` and `StoppingHighestDose` is as expected
- for `StoppingTargetProb` the logic would be that the target probability will be 0 because no active dose is available, then the message can be adjusted accordingly

0liver0815:
`StoppingTargetProb` logic changed according to comment and unit tests added.

danielinteractive (approved):

PR #594, opened 2023-06-12
587 write a concise design document for the proposed tidy generic

### LINKED ISSUE #587: Write a concise design document for the proposed `tidy()` generic
`broom`-like `tidy()`, `glance()` and `augment()` methods for `crmPack` classes would be helpful.  #323 is relevant.

Puzzled-Face:
closes #587


design/concise-tidy-design.Rmd
```diff
@@ -0,0 +1,215 @@
...
+* `tbl_<className>` is prepended to the class of the tidy tibble.
+
+### Exceptions
+*  Where a vector slot (or series of vector slots) define a range, then the naming convention described above is not followed.  Instead, columns named `min` and `max` define the extent of the range and a column named `Range` labels the range.
+
+### Please also read
+Please also read [Issue 407](https://github.com/Roche/crmPack/issues/407) regarding inconsistent slot names.
+
+## Implementation
+`tidy()` is a generic defined in `broom`.  An end-user of `crmPack` may not have installed `broom`.  Therefore, we need to define an appropriate geneic when necessary.
```
danielinteractive, line 37: I think the cleanest would be to import the `tidy` generic from the `broom` package.

design/concise-tidy-design.Rmd
```diff
@@ -0,0 +1,215 @@
...
+`CohortSizeConst` is a trivial example.
+
+```{r}
+# tidy-CohortSize ----
+
+#' @rdname tidy
+#' @aliases tidy-CohortSize
+#' @example examples/CohortSize-method-tidy.R
+#' @export
+setMethod(
```
danielinteractive, line 77: this could be a default method, defined with a signature for the top level classes, because the content is not specific

danielinteractive (commented):
Thanks @Puzzled-Face ! My first comments would be:
1) let's check what is out there in terms of "S4 class to tibble" functionality in other packages
2) top level methods would be great to have such that when the user defined new subclasses they would work out of the box, but maybe not in the most pretty way

design/concise-tidy-design.Rmd
```diff
@@ -0,0 +1,215 @@
...
+devtools::load_all()
+```
+
+## Introduction
+Here is a first draft implementation of broom-like methods for `crmPack, as suggested in [issue 323](https://github.com/Roche/crmPack/issues/323).
+
+### Principles
+The general principles I propose for tidying `crmPack` objects are as follows:
+
+* All slots that are not functions are converted to tibbles or a list of tibbles.
```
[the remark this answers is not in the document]
Puzzled-Face: I've done some searching on the web.  I've not found any packages.  There are plenty of questions on StackOverflow.  All the answers make suggestions along the lines of what I have done here: go through the slots one by one and convert each to a tibble.

See, for example, [here](https://stackoverflow.com/questions/57317958/general-way-to-transform-s4-object-to-dataframe-r) and [here](https://stackoverflow.com/questions/14604269/how-to-create-a-dataframe-of-user-defined-s4-classes-in-r).

design/concise-tidy-design.Rmd
Puzzled-Face: I agree, if you're happy to have `crmPack` Depend on `broom`.  I was trying to avoid that.

design/concise-tidy-design.Rmd
Puzzled-Face: Yes, and then that would satisfy your request for a "do it even if it's ugly" default method.

design/concise-tidy-design.Rmd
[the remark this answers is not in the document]
danielinteractive: OK cool, thanks. Can you add this information to the design doc? Will be good to be able to remind us later why we did it

design/concise-tidy-design.Rmd
danielinteractive: no worries, since it is a standard package ok to depend on it

design/concise-tidy-design.Rmd
danielinteractive: yep, good

design/concise-tidy-design.Rmd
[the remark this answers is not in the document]
Puzzled-Face: Done.

iqis:
Just air-dropped into the convo by @robertadamsbayer... Not sure if I understand the issue fully, but here's some of my thoughts based on the convo in this issue/PR. 

Generics gets registered and stored in its own namespace in a package. 
```r
environmentName(environment(broom::tidy))
```
```[1] "generics"```

If you `tidy()` doesn't use anything from the `broom::tidy()` it's not quite necessary to invite the dependency. It's safe to define your own generic named `tidy()`, so long as it's a proper generic: 

```r 
function (x, ...) 
{
    UseMethod("tidy")
}
```
It's not going to interfere with `broom::tidy()`. 

For S3 generic dispatch on an S4 object. You can directly do it because S4 objects already have the S3 class of the same name. 

```r

## The S3 Generic
wack <- function(x, ...){
  UseMethod("wack", x)
}

## Object/Instance of class <essThree>
essThreeMole <- structure(list(data = 333), 
                          class = "essThree")

## S3 Method for class <essThree>
wack.essThree <- function(x, ...){
  message(sprintf("BAAAM!! S3 Mole! The data is %s",
                 x$data))
}

## Dispatch
wack(essThreeMole)


## Class Definition of Class <essFour>
essFour <- setClass("essFour", 
                    slots = list(data = "numeric"))
## Object/Instance of class <essFour>
essFourMole <- new("essFour", data = 444)

## It already has a S3 class 
class(essFourMole)

## So just write a S3 generic 
wack.essFour <- function(x, ...){
  message(sprintf("BAAAAAAAAAAAAAAM!! S4 Mole!!!! The data is %s", 
                  x@data))
}

wack(essFourMole)
```

iqis:
@Puzzled-Face 

While I believe you can play some tricks with `purrr` to flatten nested lists into data frames, I know of no such package that readily does it with S4 objects. This appears as a niche need to me. 

You may want to reserve the thought of making it its own package for at a later lifecycle. Before that, it can be the candidate for a good starter content of, say,`RocheUtils`, which is akin to `Hmisc` or `xfun`. 

But this sounds interesting. If you haven't already implemented this part I'd like to whip up something quick. 

Now that brings about perhaps the most difficult and important question, do you plan to just name it `tidy()`?

danielinteractive:
Thanks @iqis for jumping in here and @robertadamsbayer for the connection!

Yeah sure if you would have time to put a general prototype together that would be great, because we have so many other things ongoing to finish ... I think the idea was to call it `tidy()` but I guess we could also call it something else if we have another good idea.

Puzzled-Face:
I have prototype code for most of the classes in `crmPack` and the approach I've taken is pretty generalisable, I think.  I'm happy to share what I have if you're interested.

iqis:
@Puzzled-Face 

Not sure if I can make a meaningful contribution but, if it's fine with you, why not? Let's see it.

Puzzled-Face:
@iqis Take a look at https://github.com/Roche/crmPack/pull/408.  The code is in `R/broom-methods.R` and there's an example of the use of each `tidy` method in `/design/tidy-demonstration.Rmd`.  It's still a work-in-progress, but it's pretty far advanced and is definitely usable.

PR #595, opened 2023-06-12
Closes #586

### LINKED ISSUE #586: Code in example.Rnw does not run correctly
in #581, `example.Rnw` compiles, but the code it contains produces incorrect results.  Currently, the first problem I've found is on page 99.

image

Eight pages of purple warning messages then follow.  I think the problem is related to an incomplete renaming of parameters of `DualResponsesSamplesDesign` as part of the tidy-up.  Some calls use `Effmodel`, other `eff_model` but changing `Effmodel` to `eff_model` and vice versa cause errors to be thrown.

Puzzled-Face:
…linewidth with size.

# Pull Request

Fixes #nnn

### DIFF
--- a/R/Design-methods.R
+++ b/R/Design-methods.R
@@ -3388,7 +3388,6 @@ setMethod("simulate",
             )
 
 
-
             ## what is the dose limit?
             doselimit <- maxDose(object@increments,
               data = thisData
@@ -3407,8 +3406,7 @@ setMethod("simulate",
               options = mcmcOptions
             )
 
-
-            thisSigma2 <- mean(thisEffsamples@data$sigma2)
+            thisSigma2 <- mean(thisEffsamples@data$sigma2W)
 
             thisSigma2betaW <- mean(thisEffsamples@data$sigma2betaW)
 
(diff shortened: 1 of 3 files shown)


danielinteractive:
Looks good thanks @Puzzled-Face , please can you revert the changes in the `.Rnw` file because in this R version we still need it to be `hide`, thanks

danielinteractive:
@Puzzled-Face I see:

```
--- re-building ‘crmPack-jss-paper.Rnw’ using knitr
Warning in remind_sweave(if (in.file) input, sweave_lines) :
  It seems you are using the Sweave-specific syntax in line(s) 380, 876; you may need Sweave2knitr("crmPack-jss-paper.Rnw") to convert it to knitr

Quitting from lines 381-388 [min-inf] (crmPack-jss-paper.Rnw)
Error: processing vignette 'crmPack-jss-paper.Rnw' failed with diagnostics:
object 'hide' not found
```

I think the problem is that the package uses `knitr` while before we were using `Sweave`. Would you like to try `Sweave2knitr("crmPack-jss-paper.Rnw")`  as suggested?

PR #597, opened 2023-06-13
582 approximate samples generates plot output as a side effect

### LINKED ISSUE #582: `approximate-Samples` generates plot output as a side-effect
The plot produced by `approximate` is generated as a side-effect (by `matplot`) when `verbose = TRUE`.  The plot is not part of the function's return value.  This seems sub-optimal (it is difficult for the end user to access the plot object) and makes unit testing more complicated than necessary.

Also, the online doc reads <html>
<body>

verbose | be verbose (progress statements and plot)? (default)
-- | --

</body>
</html>

which suggests that `verbose` is both misnamed and controls two options that should be independent (verbosity and production of the plot).

Suggestion:  
- modify `approximate`'s return value to be a list containing the approximated model and the plot (if requested) rather than just the approximated model
- modify `approximate` so that `verbose` controls only progress statements
- introduce a new parameter to control production of the plot. [`create_plot`, perhaps?]

Puzzled-Face:
Closes #582


examples/Sample-methods-approximate.R
```diff
@@ -1,34 +1,46 @@
...
 # should be increased for a real case.
 set.seed(94)
-posterior <- approximate(object = samples,
-                         model = model,
-                         data = data,
-                         logNormal=TRUE,
-                         control = list(threshold.stop = 0.1,
-                                        max.time = 1,
-                                        maxit = 1))
+posterior <- (approximate(
```
danielinteractive, line 34: better do this in 2 steps - first save the `approximate()` result list in a variable, and the subset it. That is much easier to read

tests/testthat/test-Samples-methods.R
```diff
@@ -23,3 +25,135 @@ test_that("names-Samples returns correct names of the parameters", {
   expect_identical(names(samples), c("alpha0", "alpha1", "beta"))
   expect_identical(names(samples2), c("a", "z"))
 })
+
+
+test_that("Approximate fails gracefully with bad input", {
```
danielinteractive, line 30:
```suggestion
test_that("approximate fails gracefully with bad input", {
```

tests/testthat/test-Samples-methods.R
```diff
@@ -23,3 +25,135 @@ test_that("names-Samples returns correct names of the parameters", {
...
+      model = model,
+      data = data,
+      points = c(1:5, "NotNumeric"),
+      control = list(threshold.stop = 0.1, max.time = 1, maxit = 1)
+    ),
+    "Assertion on 'points' failed: Must be of type 'numeric', not 'character'"
+  )
+})
+
+test_that("Approximate works correctly", {
```
danielinteractive, line 107:
```suggestion
test_that("approximate works correctly", {
```

tests/testthat/test-Samples-methods.R
```diff
@@ -23,3 +25,135 @@ test_that("names-Samples returns correct names of the parameters", {
...
+    model = model,
+    data = data,
+    create_plot = FALSE,
+    control = list(threshold.stop = 0.1, max.time = 1, maxit = 1)
+  )
+  expect_equal(length(actual1), 1)
+  expect_set_equal(names(actual1), c("model"))
+})
+
+## nolint end
```
danielinteractive, line 159: since we have cleaned up the tests, please avoid nolint start and end. All of this should be checked by the linter

vignettes/example.Rnw
```diff
@@ -37,7 +37,7 @@
 %% no indentation of paragraphs
 \noindent
 
-<<load, echo=FALSE, results=hide>>=
```
danielinteractive, line 40: I would revert the changes in this file back to the state on the main branch.

danielinteractive (commented):
Thanks John! Please see inside few comments.

examples/Sample-methods-approximate.R
Puzzled-Face: Done

tests/testthat/test-Samples-methods.R
Puzzled-Face: Done.

danielinteractive (approved):

PR #603, opened 2023-06-19
Remove commented lines and add StoppingMissingDose

### LINKED ISSUE #602: Remove commented lines added for pull request #588 from examples
This is to uncomment the lines which were commented for pull request #588 in some examples.

To quickly be able to merge the pull request, I tried to use the `\dontrun{}` feature so that examples that run into an error, were not executed. However, `\dontrun{}` is not compatible with SuperLinter, therefore I needed to comment the code that causes the error.
Now I found the problem why the examples did not run. It was quite simple. Due to the new handling of NA from `nextBest`, the new stopping rule `StoppingMissingDose` must be added to some examples. Otherwise they run into an error.

This is to roll back the changes in the examples and add the stopping rule.

Note: As I had problems to identify which code causes the error I commented more lines/files than necessary. Therefore some files where rolled back, without that it was necessary to add the stopping rule.

0liver0815:
Close #602

### DIFF
--- a/examples/Rules-method-nextBest-NextBestDualEndpoint.R
+++ b/examples/Rules-method-nextBest-NextBestDualEndpoint.R
@@ -60,6 +60,4 @@ print(dose_recommendation$plot)
 
 # Show customization of single plot.
 variant1 <- dose_recommendation$singlePlots$plot1 + xlim(0, 20)
-# nolint start
-# print(variant1)
-# nolint end
+print(variant1)
--- a/examples/Simulations-method-plot-DualSimulations.R
+++ b/examples/Simulations-method-plot-DualSimulations.R
@@ -3,6 +3,17 @@
 # Define the dose-grid
 emptydata <- DataDual(doseGrid = c(1, 3, 5, 10, 15, 20, 25, 40, 50, 80, 100))
 
+# Create some data
+data <- DataDual(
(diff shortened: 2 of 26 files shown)


danielinteractive (commented):
Thanks @0liver0815 can you also remove the nolint start and end markers where possible?

0liver0815:
> Thanks @0liver0815 can you also remove the nolint start and end markers where possible?

@danielinteractive done.

In some examples the following code is used:
```
## Deifne the increments for the dose-escalation process
## The maximum increase of 200% for doses up to the maximum of the dose specified in the doseGrid
## The maximum increase of 200% for dose above the maximum of the dose specified in the doseGrid
## This is to specified a maximum of 3-fold restriction in dose-esclation
myIncrements <- IncrementsRelative(
  intervals = c(min(data@doseGrid), max(data@doseGrid)),
  increments = c(2, 2)
)
```
or
```
# The increments (see Increments class examples).
# 200% allowable increase for dose below 300 and 200% increase for dose above 300.
my_increments <- IncrementsRelative(
  intervals = c(25, 300),
  increments = c(2, 2)
)
```
if I am not mistaken, this is the same as
```
my_increments <- IncrementsRelative(
  intervals = 0,
  increments = 2
)
```
Should I change the code?

danielinteractive:
Ah now I understand, yes please simplify as suggested 👍🏼

danielinteractive (approved):

PR #646, opened 2023-07-10
124 stop rule reporting

### LINKED ISSUE #94: Add additional statistics to the analysis of simulation output

ClaraBeckBayer:
# Pull Request

Pull request for new branch for stopping rule reporting.

Fixes #94 (partially)


R/Rules-class.R
```diff
@@ -1519,7 +1544,11 @@ StoppingCohortsNearDose <- function(nCohorts = 2L,
 #' @note Typically, end users will not use the `.DefaultStoppingCohortsNearDose()` function.
 #' @export
 .DefaultStoppingCohortsNearDose <- function() { # nolint
-  StoppingCohortsNearDose(nCohorts = 3L, percentage = 0.2)
+  StoppingCohortsNearDose(
+    nCohorts = 3L,
+    percentage = 0.2,
+    report_label = NA_character_
```
danielinteractive, line 1550:
```suggestion
    percentage = 0.2
```
not needed explicitly here, it is the default for that argument anyway. similar below for other default constructors

R/Rules-class.R
```diff
@@ -2221,14 +2534,16 @@ StoppingList <- function(stop_list, summary) {
...
 #'
 #' @param stop_list (`list`)\cr see slot definition.
-#'
+#' @param report_label (`string`) \cr see slot definition.
 #' @export
 #' @example examples/Rules-class-StoppingAll.R
 #'
-StoppingAll <- function(stop_list) {
-  .StoppingAll(stop_list = stop_list)
+StoppingAll <- function(stop_list, report_label = character(0)) {
```
danielinteractive, line 2541: why do we have here `character(0)` as default instead of `NA_character_` above?

R/Design-methods.R
```diff
@@ -1063,6 +1089,9 @@ setMethod("simulate",
       ## the reasons for stopping
       stopReasons <- lapply(resultList, "[[", "stop")
 
+      ## for dual simulations as it would fail in summary otherwise (for dual simulations reporting is not implemented)
+      stop_report <- matrix(TRUE, nrow = nsim)
```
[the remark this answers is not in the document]
ClaraBeckBayer: No, because this matrix is not supposed to contain missing values: 
image

R/Design-methods.R
```diff
@@ -398,13 +398,33 @@ setMethod("simulate",
...
+              value <- stopit_tree[1]
+              names(value) <- label
+              stopit_unpacked <- append(stopit_unpacked, value)
+              if (is.null(attr(stopit_tree, "individual"))) {
+                stopit_unpacked
+              } else {
+                append(stopit_unpacked, lapply(attr(stopit_tree, "individual"), unpack_stopit))
+              }
+            }
+          }
```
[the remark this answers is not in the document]
ClaraBeckBayer: Okay, I added it to the helpers.R file, is this the right place?

R/Rules-class.R
ClaraBeckBayer: Changed this to NA for StoopingAll and StoppingAny too.

R/Design-methods.R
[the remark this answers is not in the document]
danielinteractive: yep that works thanks!

R/helpers.R
```diff
@@ -1044,3 +1044,30 @@ h_find_interval <- function(..., replacement = -Inf) {
...
+
+#'unpack stopping rules and return list
+#'
+#'@description
+#'
+#'recursively unpack nested stopping rules logical value and label given
+#'
+#'@param stopit_tree object from simulate method
+
+h_unpack_stopit <- function(stopit_tree) {
```
danielinteractive, line 1058: can you please add unit tests covering all if/else paths for this function?

ClaraBeckBayer: Robert optimized the function so there is only one if condition left now.

R/Simulations-methods.R
```diff
@@ -977,6 +977,17 @@ setMethod("show",
         dfNames = dfNames
       )
 
+      # report stopping rules
+      # Report individual stopping rules with non-<NA> labels.
+      stop_pct <- colMeans(object@stop_report) * 100
```
danielinteractive, line 982: Can you please add tests that check whether this reporting works as expected? including potential edge cases. Since this is the main feature of this PR

danielinteractive (approved):
Just one minor comment then good to merge

R/Simulations-methods.R
```diff
@@ -979,8 +978,11 @@ setMethod("show",
 
       # report stopping rules
       # Report individual stopping rules with non-<NA> labels.
-      stop_pct <- colMeans(object@stop_report) * 100
-      stop_pct_to_print <- stop_pct[!is.na(names(stop_pct))]
+      # stop_pct <- colMeans(object@stop_report) * 100
+      # stop_pct_to_print <- stop_pct[!is.na(names(stop_pct))]
```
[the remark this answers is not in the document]
robertadamsbayer: Solved

PR #653, opened 2023-07-25
Implementing CrmPackClass class union.

Puzzled-Face:
Implementing the CrmPackClass class union.


R/class-union.R
```diff
@@ -19,7 +19,7 @@ setClass(
 
 # CrmPackClass-classunion
 
-#CohortSize
+# CohortSize
 setClassUnion(
   "CrmPackCohortSize1",
   c(
```
danielinteractive, line 25: could we automatically find those child classes instead of hardcoding them? Just thinking how we can simplify maintenance

Puzzled-Face: We could construct a tree like hierarchy, I suppose.  It could be done.  What we have now is an MVP.  But I do like the idea of automatically updating as the package evolves.  

Originally, I thought we could simply list the first ancestor in each tree: `setClassUnion("CrmPackClass", c("CohortSize", "Data", "Design", "Increments, ...)`.  But that didn't work.

Let me investigate...

Puzzled-Face:
As agreed in chat: rather than defining `CrmPackClass` as a class union, define it simply as

```
setClass("CrmPackClass")
``` 

and then make it the ultimate ancestor of all other `crmPack` classes.  For example

```
setClass("CohortSize", .... , contains = "CrmPackClass")
```

danielinteractive (commented):
Thanks @Puzzled-Face , looks good , please no TODOs in the code but instead follow on GH issue as needed

Puzzled-Face:
> Thanks @Puzzled-Face , looks good , please no TODOs in the code but instead follow on GH issue as needed

Done.  #659 already lists the new default constructors that are needed.

danielinteractive (approved):

Issue #656, opened 2023-07-26
Add additional statistics to the analysis of simulation output


0liver0815:
Add median MTD and CV to simulation output, see issue #94

PR #661, opened 2023-08-10
656 median mtd

Prerana17:
# Pull Request

Will be review by @0liver0815 initially.

Fixes #https://github.com/Roche/crmPack/issues/656

### DIFF
--- a/R/Simulations-methods.R
+++ b/R/Simulations-methods.R
@@ -634,6 +634,7 @@ setMethod("summary",
       ret <- .SimulationsSummary(
         start,
         stop_report = object@stop_report,
+        additional_stats = object@additional_stats,
         fitAtDoseMostSelected = fitAtDoseMostSelected,
         meanFit = meanFit
       )
@@ -641,7 +642,6 @@ setMethod("summary",
       return(ret)
     }
 )
-
 
 ##' Summarize the dual-endpoint design simulations, relative to given true
 ##' dose-toxicity and dose-biomarker curves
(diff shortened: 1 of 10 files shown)


R/Simulations-methods.R
```diff
@@ -979,6 +981,15 @@ setMethod("show",
 
 
 
+      cat(
```
0liver0815, line 984: Move the reporting between MostSelected and Stopping.

R/Simulations-methods.R
```diff
@@ -979,6 +981,15 @@ setMethod("show",
...
+        "Median MTD : ",
+        round(unlist(object@additional_stats)[1], 2), "\n"
+      )
+      cat(
+        "CV MTD : ",
```
0liver0815, line 989: Mean CV (MTD)
multiply by 100 and add % as unit.

0liver0815 (commented):
Initial review finished.

danielinteractive:
Hi @Prerana17 is this ready for my review? just let me know and I can do it

0liver0815 (commented):
Just minor changes to comments.

0liver0815:
To resolve the r cmd check error, I think you need to add an empty list to the other simulation methods, as e.g. done for the stopping here: https://github.com/Roche/crmPack/blob/96702347ab8b7cc09c69df80110afbd1b10bdfd6/R/Design-methods.R#L1093
To avoid empty reporting lines, I propose that the results are only printed in summary, when the list is not empty.

Prerana17:
Old PR for median and CV MTD

PR #662, opened 2023-08-15
Add grouped single/combo design prototype

danielinteractive:


Puzzled-Face:
In `/design/single_combo_grouped.Rmd`, The `prior-elicitation` chunk label is used twice. [Lines 327 and ?362.]

design/single_combo_grouped.Rmd
Puzzled-Face, line 1: Line 129 `p + facet_wrap(vars(group), nrow = 2)` fails with 

```
## Error in `combine_vars()`:
## ! At least one layer must contain all faceting variables: `group`
## ✖ Plot is missing `group`
## ✖ Layer 1 is missing `group`
## ✖ Layer 2 is missing `group`
```

`p + facet_wrap(vars(x@group), nrow = 2)` gives results that appear sensible.

Puzzled-Face, line 1: "We note that here refDose should be chosen below or equal to the lowest dose, such that the term log(dose / ref_dose) is never negative and hence the probability of DLT at any given dose will be higher for the combination than for the single agent."

I disagree.  I was once involved with an (ordinal) CRM trial in which all participants who were dosed above a particular dose reported sub toxic AEs related to bone turnover.  The trial continued with a combination of IMP + concomitant medication aimed at reducing the incidence of bone related AEs.  So, here, the deliberate intention was to create a situation where p(Tox | dose + combo) < p(Tox  | dose).

Would a more natural formulation be

logit(p | d, group) = alpha0 + delta0 + (alpha1 + delta1*(group == combo)) * log(d/d<sub>ref</sub>)?

This would allow direct evaluation of the size of the "combo effect" on both slope and intercept.

Also, my understanding is that d<sub>ref</sub> defines the point of inflexion of the logistic dose-toxicity curve on the raw probability scale.  That means it has a nice geometric interpretation.  It's the point about which the fitted curve pivots.  I believe defining d<sub>ref</sub> outside the trial's dose range may lead to instability in estimates of the model parameters.

Puzzled-Face, line 1: A revised definition of the `fit` method appears to be missing, leading to, for example,
```
fit_single <- fit(one_sample, my_model, my_data, group = factor("single", levels = c("single", "combo")))
## Error in assert_factor(group, len = length(dose), levels = c("single", : argument "group" is missing, with no default
```

Puzzled-Face:
> _ No description provided. _

General comments:

**Incomplete code?**
The PR appears incomplete to me.  The Rmd fails to knit for me.  Several errors are reported, mostly cascading from missing `group` arguments. I suspect various ffff-DataGrouped methods are missing.

**Overly focussed?**
Is this extension too focussed on combination therapy?  Could we see this as a special case of "data with covariates" or `DataCovariate`?  For example, one might want to adjust for a particular phenotype or comorbidity in the same way as use of a concomitant medication.

The class definition for the more general class could look something like this:

```
.DataCovariate <- setClass(
  Class = "DataCovariates",
  slots = c(
    covariates = "list"
  ),
  prototype = prototype(
    covariates = list()
  ),
  contains = "Data"
)

DataCovariate <- function(covariates,
                        ...) {
  d <- Data(...)
  .DataCovariates(
    d,
    covariates = covariates
  )
}
```

So that combination therapy could be handled with 

```
comboData <- DataGrouped(..., covariates=list(group=factor(rep(c("single", "combo"), each=5), levels = c("single", "combo")))
```

The disadvantage of this approach is that users would need to provide custom functions for the `datamodel` slot and `dose`, `prob` and other methods.

The two approaches could be combined by making the proposed `DataGrouped` class to be a sub-class of `DataCovariates`.  Thus, end users have a ready-rolled class for a common use case (the most common use case) as well as the flexibility of creating their own custom models if they feel the effort is worthwhile.

design/single_combo_grouped.Rmd
danielinteractive: I just checked and it works for me, I think with installing the package from this branch this should disappear but you can also try to update your `ggplot2` package in addition

danielinteractive: similar as above this should disappear upon installing branch version of package

danielinteractive: Thanks, renamed to `delta{0,1}`!

Plus added some more thoughts on choice of the ref dose.

tests/testthat/test-Samples-methods.R
Puzzled-Face, line 1: Line 215: Perhaps also, for completeness,

```
  expect_equal(nrow(actual), length(my_data@doseGrid))
  expect_numeric(actual, any.missing = FALSE, lower = 0)
```

Puzzled-Face (approved):

design/single_combo_grouped.Rmd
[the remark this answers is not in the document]
danielinteractive: It is here: 
image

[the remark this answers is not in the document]
danielinteractive: done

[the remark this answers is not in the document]
danielinteractive: good point, I now have mono everywhere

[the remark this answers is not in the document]
danielinteractive: hm good point. changing same_dose to mean the minimum of both doses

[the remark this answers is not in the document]
danielinteractive: ok done (not that important here though because this won't be package code)

tests/testthat/test-Samples-methods.R
danielinteractive: done

R/Simulations-methods.R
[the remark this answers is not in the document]
danielinteractive: done

PR #663, opened 2023-08-16
New commit for additional statistics

Prerana17:
closes #https://github.com/Roche/crmPack/issues/656


R/Design-methods.R
```diff
@@ -417,6 +417,19 @@ setMethod("simulate",
           data = thisData
         )
 
+        # Get the MTD estimate from the samples.
+        sample_mtd <- dose(
+          mean(object@nextBest@target),
+          model = object@model,
+          samples = thisSamples
+        )
```
danielinteractive, line 425: Why is the MTD determined like this? I am wondering because we use `thisDose` elsewhere here as the MTD (see below)

R/Design-methods.R
```diff
@@ -417,6 +417,19 @@ setMethod("simulate",
...
+
+        # Create list with median MTD and CV.
+        additional_stats <- list(
```
danielinteractive, line 428: this is nice I think, however I wonder if we can make this more flexible for the user - e.g. how about accepting a list of functions as part of the `simulate()` call and then the user can define `median`, `cv` functions etc. to apply to `target_dose_samples`?

danielinteractive (commented):
Thanks @Prerana17 and @0liver0815 please see comments below

danielinteractive:
@Prerana17 @ClaraBeckBayer @0liver0815 I am assuming this is the PR moving forward now and not https://github.com/Roche/crmPack/pull/661 ? or is that not the case? ... please clarify

R/Design-methods.R
0liver0815: Thanks for the suggestion. If I am not mistaken it is the estimated MTD per sample, i.e. the calculated dose that corresponds to to the target (or the mean of the interval if target is an interval) for a specific sample. But I see the point that the dose is the exact dose and not likely to be a dose from the dosegrid. Therefore I think the proposal makes sense to rename it to `target_dose_samples`.

R/Design-methods.R
```diff
@@ -417,6 +417,19 @@ setMethod("simulate",
...
+          median_mtd = median(sample_mtd),
+          cv_mtd = mad(sample_mtd) / median(sample_mtd)
+        )
+
         ## return the results
         thisResult <-
           list(
```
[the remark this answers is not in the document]
0liver0815: With the explanation above it is probably ok now? Or should we rename `median_mtd` to avoid any confusion with the selected dose from the dose grid at the end of the trial? E.g. `sample_median_mtd`?

R/Design-methods.R
0liver0815: Great proposal. I really like that idea. This will add much flexibility e.g. if someone want to see the mean instead of the median etc.. We just have to evaluate, if it is straight forward to implement.

0liver0815:
After talking to @Prerana17, the other PR is closed and the changes that @ClaraBeckBayer added, that were not part of this PR will be added here, so that this PR is up to date. @danielinteractive I think a review makes only sense for you, if the all checks completed without error.

danielinteractive:
Thanks @0liver0815 and @Prerana17 for the clarification :) all good. I did initial review here, please have a look since that could change structure a bit.

R/Design-methods.R
Prerana17: @danielinteractive @0liver0815 Sure, as the next step, I will try to implement the functions where user can define median, cv, mean etc. functions.

R/Design-methods.R
```diff
@@ -209,7 +209,7 @@ setMethod("simulate",
              truth, args = NULL, firstSeparate = FALSE,
              mcmcOptions = McmcOptions(),
              parallel = FALSE, nCores =
-               min(parallel::detectCores(), 5),
+               min(parallel::detectCores(), 5), derive = NULL,
```
0liver0815, line 212: I think the default could be an epmty list.

```suggestion
               min(parallel::detectCores(), 5), derive = list(),
```

R/Design-methods.R
```diff
@@ -4636,6 +4677,21 @@ setMethod("simulate",
...
+          model = object@model,
+          samples = thisSamples
+        )
+
+        # Create a function for additional statistical summary.
+        if (is.null(derive)) {
+          additional_stats <- list()
+        } else {
+          additional_stats <- lapply(derive, function(f) f(target_dose_samples))
+        }
```
0liver0815, line 4693: Same here.

R/Simulations-class.R
```diff
@@ -125,7 +127,9 @@ GeneralSimulations <- function(data,
         ),
       stop_report = matrix(TRUE, nrow = 2),
       stop_reasons =
-        list("A", "A")
+        list("A", "A"),
+      additional_stats =
+        list(1, 1)
```
0liver0815, line 132: Probably this needs to be functions instead of (1,1).

examples/Simulations-class-DualSimulations.R
```diff
@@ -27,13 +27,16 @@ stop_report <- matrix(c(TRUE, FALSE), nrow = 2)
 
 stop_reasons <- list("A", "B")
 
+additional_stats <- list(1, 1)
```
0liver0815, line 30: Same as above. Functions instead of (1,1).

R/Simulations-class.R
```diff
@@ -107,6 +107,7 @@ GeneralSimulations <- function(data,
 #' @slot fit (`list`)\cr final fits
 #' @slot stop_reasons (`list`)\cr stopping reasons for each simulation run
 #' @slot stop_report matrix of stopping rule outcomes
+#' @slot additional_stats list of addtional statistical summary
```
0liver0815, line 110:
```suggestion
#' @slot additional_stats list of additional statistical summary
```

R/Design-methods.R
0liver0815: I think you need to add `derive = list()`to the function as done in line 212 above.

R/Design-methods.R
```diff
@@ -209,7 +209,7 @@ setMethod("simulate",
              truth, args = NULL, firstSeparate = FALSE,
              mcmcOptions = McmcOptions(),
              parallel = FALSE, nCores =
-               min(parallel::detectCores(), 5),
+               min(parallel::detectCores(), 5), derive = list(),
```
0liver0815, line 212: `derive` needs to be added to the to the description above
`@param derive A list of functions that derive statistics from the target dose samples.`

R/Simulations-class.R
```diff
@@ -153,6 +157,12 @@ GeneralSimulations <- function(data,
           "stop_report must be a matrix of mode logical in which the number of rows equals the number of simulations
       and which must not contain any missing values"
         )
+
+        o$check(
+          identical(length(object@additional_stats), nSims),
+          "additional_stats must have same length as the number of simulations"
+        )
```
danielinteractive, line 164: also check that it has names?

R/Simulations-methods.R
```diff
@@ -978,14 +978,23 @@ setMethod("show",
...
       )
 
+      # Report results of additional statistics summary
+
+      if (length(list()) > 0) {
+        summary_stat_op <- unlist(object@additional_stats)
+
+        cat(
+          "Results of Additional Statistical Calculation : \n",
+          paste(names(summary_stat_op), ":", round(summary_stat_op), "\n")
```
danielinteractive, line 994: ok as a first version, just wondering if we should follow up in another PR to make this printing more flexible

danielinteractive (commented):
thanks @Prerana17 , nice work! please see comments below, can you also add a `NEWS` entry?

R/Design-methods.R
```diff
@@ -417,6 +420,18 @@ setMethod("simulate",
...
+
+        target_dose_samples <- dose(
+          mean(object@nextBest@target),
+          model = object@model,
+          samples = thisSamples
+        )
+
+        # Create a function for additional statistical summary.
+
+        additional_stats <- lapply(derive, function(f) f(target_dose_samples))
```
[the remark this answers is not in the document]
Prerana17: It's throwing error. So I kept same.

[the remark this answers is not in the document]
danielinteractive: yep makes sense actually

danielinteractive (approved):

PR #665, opened 2023-08-25
664: Add `DataGrouped` class

### LINKED ISSUE #664: Add `DataGrouped` class
See https://github.com/Roche/crmPack/blob/main/design/single_combo_grouped.Rmd#L24.

To do:

- [x] Add class
    - [x] doc
	- [x] with validation function
	- [x] with unit tests
- [x] Add `groupData()` function
	- [x] doc
	- [x] with unit tests
- [x] We don't need an `update` method actually since simulation starts with separate mono and combo objects
- [x] Add `plot` method - also skipped for now. Both could be revisited later once we really go to use this in practice

danielinteractive:
closes #664


tests/testthat/helper-data.R
Puzzled-Face, line 1: A thought: The new `.DefaultXXXX()` constructors could probably replace the utility functions in the future.

R/Data-class.R
Puzzled-Face, line 1: Please add a `.DefaultDataGrouped` constructor.

Puzzled-Face (changes requested):

R/Data-class.R
danielinteractive: thx, done

tests/testthat/helper-data.R
danielinteractive: maybe, but typically the default constructors produce pretty "empty" objects while for tests we typically need more "realistic" objects

Puzzled-Face (approved):

PR #672, opened 2023-09-11
Test fixes - adding handling for empty expression `x[i, ]` edge case

dgkf:
# Pull Request

To address test failures specifically when `covr` is used, this updates the way traces are stripped from model specification code so that snippets with an empty expression (`x[i, ]`) don't trigger the "argument missing with no default" error.

A very curious behavior of R here!


danielinteractive (approved):
awesome, thanks @dgkf !

Issue #671, opened 2023-09-11
Clean up `Design-methods.R`


danielinteractive:
To do:

Clean up in below steps by separate PRs (to avoid too large PRs).
The functions are currently huge and spaghetti code ... if we can refactor them with smaller helper functions that ideally can be reused across simulation & examine methods that would be really good.

When somebody takes a task please put your name next to it (edit this description) so we don't duplicate any work :) thanks!

- [x] `setSeed` and `getResultList` @0liver0815 
- [x] Prototyping for shorter code for `simulate` with helper functions -> @ClaraBeckBayer @robertadamsbayer on branch https://github.com/Roche/crmPack/tree/671_simulate_for_Design
	- [x] see also relatively new `simulate` method for `DesignGrouped` - https://github.com/Roche/crmPack/blob/main/R/Design-methods.R#L4635
- [x] helper functions:
	- [x] https://github.com/Roche/crmPack/issues/738 @ClaraBeckBayer @robertadamsbayer --> PR sent
	- [x] https://github.com/Roche/crmPack/issues/739 @ClaraBeckBayer @robertadamsbayer 
	- [x] https://github.com/Roche/crmPack/issues/744 --> possible to combine with helper function for https://github.com/Roche/crmPack/blob/main/R/Design-methods.R#L4635 ? @ClaraBeckBayer @robertadamsbayer 

Let's first:
- [ ] increase coverage, see coverage report accessible from the package website and -> @robertadamsbayer on branch https://github.com/Roche/crmPack/tree/671_simulate_for_Design
- [ ] make tests tolerant towards small numeric deviations
- [ ] first CRAN release can go without fully refactored methods, otherwise it will take too long

don't continue below yet, this will be done when we have the first version of the helper functions in place in the package:
- [x] `simulate` for `Design` (https://github.com/Roche/crmPack/blob/main/R/Design-methods.R#L200) 
- [ ] `simulate` for `RuleDesign` (https://github.com/Roche/crmPack/blob/main/R/Design-methods.R#L514) 
- [ ] `simulate` for `DualDesign` (https://github.com/Roche/crmPack/blob/main/R/Design-methods.R#L698) 
- [ ] `examine` generic and method for `Design` (https://github.com/Roche/crmPack/blob/main/R/Design-methods.R#L1147)
- [ ] `examine` for `DADesign` (https://github.com/Roche/crmPack/blob/main/R/Design-methods.R#L1508)
[...]

PR #676, opened 2023-09-14
Refactor setSeed and getResultList

0liver0815:
Closes a part of #671


R/Design-methods.R
```diff
@@ -1,165 +1,151 @@
...
+## set_seed ----
+
+#' Helper function to set and save the RNG seed
+#'
+#' @description `r lifecycle::badge("stable")`
+#'
+#' This code is basically copied from `stats:::simulate.lm`.
+#'
+#' @param seed an object specifying if and how the random number generator
+#' should be initialized (\dQuote{seeded}). Either `NULL` (default) or an
```
danielinteractive, line 21:
```suggestion
#' should be initialized ("seeded"). Either `NULL` (default) or an
```

R/Design-methods.R
```diff
@@ -1,165 +1,151 @@
...
+
+## set_seed ----
+
+#' Helper function to set and save the RNG seed
+#'
+#' @description `r lifecycle::badge("stable")`
+#'
+#' This code is basically copied from `stats:::simulate.lm`.
+#'
+#' @param seed an object specifying if and how the random number generator
```
danielinteractive, line 20:
```suggestion
#' @param seed (`integer` or `NULL`)\cr specifies if and how the random number generator
```

R/Design-methods.R
```diff
@@ -1,165 +1,151 @@
...
+#' should be initialized (\dQuote{seeded}). Either `NULL` (default) or an
+#' integer that will be used in a call to [`set.seed`] before
```
danielinteractive, line 22:
```suggestion
#' integer that will be used in a call to [set.seed()] before
```

R/Design-methods.R
```diff
@@ -1,165 +1,151 @@
...
+#' simulating the response vectors. If set, the value is saved as the
+#' `seed` slot of the returned object. The default, `NULL` will
+#' not change the random generator state.
+#' @return The RNGstate will be returned, in order to call this function
```
danielinteractive, line 26:
```suggestion
#' @return The integer vector containing the random number generate state will be returned, in order to call this function
```

R/Design-methods.R
```diff
@@ -1,165 +1,151 @@
...
+#' with this input to reproduce the obtained simulation results.
+#'
+#' @export
+#' @keywords programming
+set_seed <- function(seed = NULL) {
   if (!exists(".Random.seed", envir = .GlobalEnv, inherits = FALSE)) {
     runif(1)
   }
 
   if (is.null(seed)) {
```
danielinteractive, line 36:
```suggestion
rng_state <- if (is.null(seed)) {
```

R/Design-methods.R
```diff
@@ -1,165 +1,151 @@
...
-    RNGstate <- get(".Random.seed", envir = .GlobalEnv)
+    rng_state <- get(".Random.seed", envir = .GlobalEnv)
```
danielinteractive, line 37:
```suggestion
    get(".Random.seed", envir = .GlobalEnv)
```

R/Design-methods.R
```diff
@@ -1,165 +1,151 @@
...
   } else {
-    R.seed <- get(".Random.seed", envir = .GlobalEnv)
-    ## make sure R.seed exists in parent frame:
-    assign("R.seed", R.seed, envir = parent.frame())
+    r_seed <- get(".Random.seed", envir = .GlobalEnv)
+    # Make sure r_seed exists in parent frame.
+    assign("r_seed", r_seed, envir = parent.frame())
     set.seed(seed)
-    RNGstate <- structure(seed, kind = as.list(RNGkind()))
+    rng_state <- structure(seed, kind = as.list(RNGkind()))
```
danielinteractive, line 43:
```suggestion
    structure(seed, kind = as.list(RNGkind()))
```
and move the `do.call()` above this so that this is the last value

R/Design-methods.R
```diff
@@ -1,165 +1,151 @@
...
-      list(quote(assign(".Random.seed", R.seed, envir = .GlobalEnv))),
+      list(quote(assign(".Random.seed", r_seed, envir = .GlobalEnv))),
       envir = parent.frame()
     )
-    ## here we need the R.seed in the parent.frame!
+    # Here we need the r_seed in the parent.frame!
   }
 
-  return(RNGstate)
+  return(rng_state)
```
danielinteractive, line 51:
```suggestion
  rng_state
```

R/Design-methods.R
```diff
@@ -1,165 +1,151 @@
...
+set_seed <- function(seed = NULL) {
   if (!exists(".Random.seed", envir = .GlobalEnv, inherits = FALSE)) {
     runif(1)
   }
 
   if (is.null(seed)) {
-    RNGstate <- get(".Random.seed", envir = .GlobalEnv)
+    rng_state <- get(".Random.seed", envir = .GlobalEnv)
   } else {
-    R.seed <- get(".Random.seed", envir = .GlobalEnv)
```
danielinteractive, line 53: also assert here that `seed` argument is an integer vector here using `assert_integer()`

R/Design-methods.R
```diff
@@ -1,165 +1,151 @@
...
-    ## make sure R.seed exists in parent frame:
-    assign("R.seed", R.seed, envir = parent.frame())
+    r_seed <- get(".Random.seed", envir = .GlobalEnv)
+    # Make sure r_seed exists in parent frame.
+    assign("r_seed", r_seed, envir = parent.frame())
```
danielinteractive, line 41: improvement idea: let's call it
```suggestion
    assign(".r_seed", r_seed, envir = parent.frame())
```
so that it is hidden from the user

R/Design-methods.R
```diff
@@ -1,165 +1,151 @@
...
     set.seed(seed)
-    RNGstate <- structure(seed, kind = as.list(RNGkind()))
+    rng_state <- structure(seed, kind = as.list(RNGkind()))
     do.call("on.exit",
-      list(quote(assign(".Random.seed", R.seed, envir = .GlobalEnv))),
+      list(quote(assign(".Random.seed", r_seed, envir = .GlobalEnv))),
```
danielinteractive, line 45:
```suggestion
      list(quote(assign(".Random.seed", .r_seed, envir = .GlobalEnv))),
```
so we need to change this here too

R/Design-methods.R
```diff
@@ -1,165 +1,151 @@
...
+#' single parameter the iteration index.
+#' @param nsim number of simulations to be conducted.
+#' @param vars names of the variables.
+#' @param parallel should the simulation runs be parallelized across the
+#' clusters of the computer?
+#' @param n_cores how many cores should be used for parallel computing?
+#' @return The list with all simulation results (one iteration corresponds
+#' to one list element).
+#'
+#' @importFrom parallel makeCluster clusterEvalQ clusterExport clusterCall parLapply stopCluster
```
danielinteractive, line 73: this is a bit of a change, now we only want to import a single function from a package, and we always use `parallel::makeCluster()` i.e. with the namespace specification.
```suggestion
#' @importFrom parallel makeCluster 
```

R/Design-methods.R
```diff
@@ -1,165 +1,151 @@
...
 }
 
+## get_result_list ----
+
+#' Helper function to obtain simulation results list
+#'
+#' @description `r lifecycle::badge("stable")`
+#'
+#' The function `fun` can use variables that are visible to itself.
+#' The names of these variables have to given in the vector `vars`.
```
danielinteractive, line 61:
```suggestion
#' The names of these variables have to be given in the vector `vars`.
```

R/Design-methods.R
```diff
@@ -1,165 +1,151 @@
...
+#'
+#' @param fun the simulation function for a single iteration, which takes as
```
danielinteractive, line 63: can you also add the type of the arguments , e.g. here
```suggestion
#' @param fun (`function`)\cr the simulation function for a single iteration, which takes as
```

danielinteractive (commented):
Thanks a lot @0liver0815 , much appreciated, please see few comments below

R/Design-methods.R
0liver0815: One question: Do we want to have seed as integer? set.seed() accepts any number and interpret it as integer, so we should probably accept the same?

R/Design-methods.R
0liver0815: Changed.

R/Design-methods.R
0liver0815: Changed.

R/Design-methods.R
0liver0815: Here it may not be necessary as .r_seed and r_seed are identical.

0liver0815: But changed.

R/Design-methods.R
danielinteractive: I would say here `integer` (that is also in the sentence afterwards) and we can silently convert numeric to integer inside (with `safeInteger()`)

R/Design-methods.R
danielinteractive: so we just need a single import of something, it does not have to be `makeCluster` though, to make `R CMD check` happy

R/Design-methods.R
0liver0815: OK. Just tested `safeInteger()` to convert a numeric value to an integer. The function did not convert e.g. `1.1` to `1`.  I am not sure how the function works as I expect.

0liver0815: I think safeInteger() only converts values that are `is.wholenumber(x)`. I will convert seed directly with `as.integer` here.

danielinteractive:
@0liver0815 seems there are some failing unit tests:
image
let me know if you need help with that

R/Design-methods.R
```diff
@@ -18,37 +18,40 @@ NULL
...
+    do.call(
+      "on.exit",
+      list(quote(assign(".Random.seed", .r_seed, envir = .GlobalEnv))),
       envir = parent.frame()
     )
-    # Here we need the r_seed in the parent.frame!
+    structure(seed, kind = as.list(RNGkind()))
   }
-
-  return(rng_state)
```
danielinteractive: we should still return this, just without explicit `return` statement, so
just `rng_state` here is fine. Alternatively can also omit the `rng_state` variable assignment in line 39 altogether, then the last value is the thing from the if/else outcome, so what we want

0liver0815:
> @0liver0815 seems there are some failing unit tests: image let me know if you need help with that

Thanks for offering. I finally found the problem. I changed `RNGkind` to   `RNGkind("Super-Duper")`
https://github.com/Roche/crmPack/blob/6c6fd66085516e3ee56affeba9ce11b47cdfd175/tests/testthat/test-Design-methods.R#L14
 in one of the tests, which had downstream interference with some other unit tests. I reset back to `RNGkind("default")` at the end of the unit test, which solved the problem.

danielinteractive (approved):

R/Design-methods.R
danielinteractive: and I guess it is actually returned implicitly right now, but it is not easily readable as that...

0liver0815: I removed `rng_state` variable assignment.

Issue #683, opened 2023-09-22
Clean up


danielinteractive:
To do:

Clean up in below steps by separate PRs (to avoid too large PRs).

- Please first check if a helper function is really used in the package, and whether it could be replaced by a standard function from `checkmate` etc.
- If it makes sense, please move helper functions to separate helper files such as `helpers_data.R` etc. or create a new such separate helper file. 
- Transition to snake_case and prefix internal helper function names also with `h_` to indicate that they are helper functions.
- Add tests everywhere, add examples for user facing functions

When somebody takes a task please put your name next to it (edit this description) so we don't duplicate any work :) thanks!

- [x] `matchTolerance`
- [x] `%~%` [John Kirkpatrick #688]
- [x] `noOverlap` John Kirkpatrick #694]
- [x] `is.scalar` (e.g. this one I am pretty sure we can replace by the `checkmate` function `test_scalar()`) [John Kirkpatrick #686]
- [x] `is.bool` (-> `test_flag()`) [John Kirkpatrick #686]
- [x] `is.wholenumber` (`-> `test_number()`) [John Kirkpatrick #693]
- [x] `safeInteger` [John Kirkpatrick #693]
- [x] `is.probability` (remove because we have already https://github.com/Roche/crmPack/blob/main/R/checkmate.R#L117) 
          [**Dimitris** #692]
- [x] `is.range` (remove I guess) 
          [**Dimitris** #692]
- [x] `is.probRange` (remove because we have already https://github.com/Roche/crmPack/blob/main/R/checkmate.R#L159)
          [**Dimitris** #692]
- [x] `logit` and `probit`[Nothing to do: see discussion below.]
- [x] `crmPackExample` and `crmPackHelp` -> this is a user facing function, so no `h_` prefix here [Nothing to do here.  See below.]
- [x] `plot.gtable` and `print.gtable` -> needed? [John Kirkpatrick #696]
- [x] `dinvGamma`, `pinvGamma`, `qinvGamma`, `rinvGamma` -> replace by using other package's function? [No action: see below]
- [x] `myBarplot`
- [x] `multiplot` [John Kirkpatrick #699]

Puzzled-Face:
`%~%`  appears to be used only in the example documenting the function...

Puzzled-Face:
`multiplot` appears only to be used in `design/devel/rolling-crm-Simona.Rmd` and its functionality exists in other packages, such as `cowplot` and `ggpubr`.  Suggest removal.

[...]

PR #687, opened 2023-10-03
replace/remove 'is.probability'/'is.range'/'is.probRange'  with applicable function from additional 'checkmate' assertions

dimitriskontosBAY:
# Pull Request

Fixes one task from #683


Puzzled-Face:
Dimitris, you'll have conflicts with the changes I made to remove `is.bool` and `is.scalar`.  I've just merged the changes....

PR #691, opened 2023-10-05
Remove `safeInteger` and `is.wholenumber`

Puzzled-Face:
The initial edit is straightforward.  However, its implications are huge because almost every file in the package is impacted by the change: `checkmate::assert_integer`, for example, requires the *type* of its input is integer, not just that the values provided are integers.  Thus,

```
my_cohort_size <- CohortSizeConst(3)
```

is now invalid and must be changed to

```
my_cohort_size <- CohortSizeConst(3L)
```

Also, I have fixed a bug in `h_get_mcmc_options`: you can't set the RNG seed without also providing the RNG name.  I've also written some tests to check the fix [test-helper-McmcOptions.R`.  [Although since the helper is a helper only for tests, I'm not sure that's appropriate...]

As a result of the change to `h_get_mcmc_options`, two tests in `test-Model-methods.R` now fail.  They both concern `doseFunction` for `ModelPseudo` [lines 110 and 287].  The problem is that the RNG name no longer matches.  I suspect the fix is straightforward, but I'm going boss eyed and probably missing something obvious...


danielinteractive:
Thanks @Puzzled-Face for the proposal,
just for next time, it would probably be more efficient to first align on the idea and then implement it. 
I think we should not get less user friendly here. However, maybe using `assert_count()` as well as when needed `assert_integerish()` could help to keep user friendliness and still be more streamlined internally (where you can then just use `as.integer()` in constructor etc. because we have asserted this before)?

Puzzled-Face:
> Thanks @Puzzled-Face for the proposal, just for next time, it would probably be more efficient to first align on the idea and then implement it. I think we should not get less user friendly here. However, maybe using `assert_count()` as well as when needed `assert_integerish()` could help to keep user friendliness and still be more streamlined internally (where you can then just use `as.integer()` in constructor etc. because we have asserted this before)?

I was afraid you'd say that!  I did consider that approach, but it was inconsistent with what we did when changing `is.bool` and `is.scalar`.  The changes in this PR are widespread, almost universal, but very simple.  The biggest issue (apart from breaking end-user code) is likely to be resolving conflicts with other ongoing PRs if and when this PR is merged to main.

If we go down that route, I would suggest abandoning the edits in this PR and starting over.  (I think it would be better to use `check_count` and `check_integer` and issue a warning when necessary rather than using `assert_count` and `assert_integer` as well as `assert_integerish`...)

danielinteractive:
> If we go down that route, I would suggest abandoning the edits in this PR and starting over. (I think it would be better to use `check_count` and `check_integer` and issue a warning when necessary rather than using `assert_count` and `assert_integer` as well as `assert_integerish`...)

Yeah I think that makes sense. For me the important feature to keep is that the user must be able to continue supplying e.g. `3` to the constructors etc. where internally they can be safely converted to `3L`. Then internally refactoring this to avoid custom assertion / helper functions where it makes sense is an improvement.

Puzzled-Face:
Closed following discussion with Daniel.  The new approach is to use `xxxx_integerish` to allow more tolerance of existing end-user code.

PR #692, opened 2023-10-05
683 replace/remove is.probability/is.range/is.prob range with applicable additional checkmate assertions

dimitriskontosBAY:
# Pull Request

Addresses the following taks from #683
- is.probability 
- is.range 
- is.probRange


R/Rules-methods.R
```diff
@@ -3017,7 +3017,7 @@ setMethod("stopTrial",
       prob_target <- stopping@prob_target
 
       ## checks
-      stopifnot(is.probability(prob_target))
+      stopifnot(test_probability(prob_target))
```
danielinteractive, line 3020: please use instead `assert_probability(prob_target)`, to avoid the `stopifnot`

danielinteractive (commented):
Thanks @dimitriskontosBAY , please see below

R/Rules-methods.R
```diff
@@ -3135,7 +3135,7 @@ setMethod("stopTrial",
...
```
dimitriskontosBAY, line 3138: @danielinteractive -> as above should I use instead assert_probability(prob_target), to avoid the stopifnot?

R/Rules-methods.R
dimitriskontosBAY: Updates completed, passed all checks

R/Rules-methods.R
dimitriskontosBAY: Updates completed, passed all checks

R/fromQuantiles.R
```diff
@@ -235,8 +235,8 @@ Quantiles2LogisticNormal <- function(dosegrid,
 ##' @keywords programming
 getMinInfBeta <- function(p, q) {
   stopifnot(
```
danielinteractive, line 237: here we should also use `assert_probability` and avoid `stopifnot`

R/fromQuantiles.R
```diff
@@ -311,10 +311,10 @@ MinimalInformative <- function(dosegrid,
   nDoses <- length(dosegrid)
   stopifnot(
     !is.unsorted(dosegrid, strictly = TRUE),
-    is.probability(threshmin, bounds = FALSE),
-    is.probability(threshmax, bounds = FALSE),
-    is.probability(probmin, bounds = FALSE),
-    is.probability(probmax, bounds = FALSE)
+    test_probability(threshmin, bounds = FALSE),
```
danielinteractive, line 314: here too

danielinteractive (commented):
almost there!

R/fromQuantiles.R
dimitriskontosBAY: Updates completed

R/fromQuantiles.R
dimitriskontosBAY: Updates completed

R/fromQuantiles.R
```diff
@@ -87,7 +89,6 @@ Quantiles2LogisticNormal <- function(dosegrid,
     identical(length(upper), nDoses),
     all(lower < median),
     all(upper > median),
-    is.probability(level, bounds = FALSE),
```
dimitriskontosBAY, line 90: also here used assert_probability and avoid stopifnot.

danielinteractive (approved):
thanks @dimitriskontosBAY !

PR #697, opened 2023-10-09
Remove plot.gtable and print.gtable

Puzzled-Face:
Part of #683.


danielinteractive (approved):

danielinteractive:
@Puzzled-Face I think we just need to revert this PR to fix #716 . I think we should not use any `grid.draw()` explicitly elsewhere

Puzzled-Face:
We could retain this PR (thus removing `print.gtable` and `plot.gtable` - a small but reasonable simplification) and revise #716 to replace new and existing calls to `grid.draw()` with `gridExtra::grid.arrange()`.  That seems to give us the best of both worlds...

Puzzled-Face:
"Sorry, this pull request couldn’t be reverted automatically. It may have already been reverted, or the content may have changed since it was merged."

danielinteractive:
Yeah I guess you might need to make a "manual revert" of this PR, by adding the contents this PR deleted in a new PR

PR #701, opened 2023-10-11
add feature needed to stop mono arm when combo arm stops

danielinteractive:


Puzzled-Face:
FYI

  ```
Examples with CPU (user + system) or elapsed time > 5s
                                               user system elapsed
   simulate-DesignGrouped-method             33.515  0.394  33.585
   summary-PseudoDualSimulations-method       9.448  1.375   9.297
   show-PseudoDualSimulationsSummary-method   9.325  1.407   9.187
```

The `simulate-DesignGrouped-method` example is the longest running of all examples, but factor of more than 3.  Can its run time be reduced?

Perhaps change `nsim = 4` to `nsim = 2` or `nsim = 1`?

**Review by Puzzled-Face — APPROVED** [review]

**Comment by danielinteractive** [comment]
Thanks @Puzzled-Face - one question, when running through the example I noticed that the lines
```r
plot(mono_sims_sum)
plot(combo_sims_sum)
```
no longer produce the plots as before - could that be due to the removed `multiplot` thing? would you mind having a look?

Puzzled-Face:
> Thanks @Puzzled-Face - one question, when running through the example I noticed that the lines
> 
> ```r
> plot(mono_sims_sum)
> plot(combo_sims_sum)
> ```
> 
> no longer produce the plots as before - could that be due to the removed `multiplot` thing? would you mind having a look?

I missed this.  Apologies.  Will take a look.

PR #703, opened 2023-10-11
Refactor myBarplot

Puzzled-Face:
Part of #683.


R/helpers_simulations.R
```diff
@@ -0,0 +1,33 @@
...
+#' @param x vector of samples
+#' @param description xlab string
+#' @param xaxisround rounding for xaxis labels (default: 0, i.e. integers will
+#' be used)
+#'
+#' @return the ggplot2 object
+#'
+#' @keywords internal
+#' @importFrom ggplot2 ggplot geom_histogram aes xlab ylab xlim
+#' @example examples/barplot_percentages.R
```
danielinteractive, line 12: because it is internal, we should not have examples (they will fail because not exported) and we should also consider prefixing the name with `h_`

examples/barplot_percentages.R
```diff
@@ -0,0 +1,4 @@
+crmPack:::barplot_percentages(
```
danielinteractive, line 1: just remove the example, because this is not acceptable for CRAN

tests/testthat/test-helpers-simulations.R
```diff
@@ -0,0 +1,36 @@
+test_that("barplot_percentages produces consistent results", {
+  set.seed(504409)
+
+  plot <- crmPack:::barplot_percentages(
```
danielinteractive, line 4: no need to use this here, because tests have access to internal functions
```suggestion
  plot <- barplot_percentages(
```

danielinteractive (commented):
thanks @Puzzled-Face , please see below comments

danielinteractive (commented):
thanks @Puzzled-Face , please see below comments

danielinteractive (approved):

Issue #705, opened 2023-10-12
Replace `matchTolerance` with `assert_equal`


Puzzled-Face:
Write an `assert_equal` `checkmate`-like function and use it as a replacement for `matchTolerance`.

See #683 for details.

Puzzled-Face:
The title of this issue is misleading.  An `assert_equal` function is not a replacement for `matchTolerance`.  Given a value, a list/vector and a tolerance, `matchTolerance` selects the (first) match to the value from the vector/list that is within tolerance.  It does not confirm that the value(s) in the first argument are all present in the second argument, allowing for given tolerance.

Therefore, it would be better simply to rename `matchTolerance` to `match_within_tolerance`.

PR #706, opened 2023-10-13
Refactoring matchTolerance

Puzzled-Face:
See #683 and #705 for details


R/checkmate.R
```diff
@@ -21,6 +21,107 @@
...
+#' @description `r lifecycle::badge("experimental")`
+#' Elements of `...` must be numeric vectors or scalars.
+#'
+#' This function performs an element-by-element comparison of the first object
+#' provided in `...` with every other object in `...` and returns `TRUE` if all
+#' comparisons are equal within a given tolerance and `FALSE` otherwise.
+#'
+#' @param ... (`numeric`)\cr vectors to be compared
+#' @param tol (`numeric`)\cr the maximum difference to be tolerated when
+#' judging equality
```
danielinteractive, line 37:
```suggestion
#' @param ... (`numeric`)\cr vectors to be compared.
#' @param tol (`numeric`)\cr the maximum difference to be tolerated when
#' judging equality.
```

R/checkmate.R
```diff
@@ -21,6 +21,107 @@
...
+#'
+#' @note If there are any missing or infinite values in `...`, this function
+#'   returns `FALSE`, regardless of the values of other elements in `...`.
+#'
+#' @note If elements in `...` are not all of the same length, `FALSE` is returned.
+#'
+#' @return TRUE if all element-by-element differences are less than `tolerance`
```
danielinteractive, line 44:
```suggestion
#' @return `TRUE` if all element-by-element differences are less than `tolerance`
```

R/checkmate.R
```diff
@@ -21,6 +21,107 @@
...
+#' @seealso [`assertions`] for more details.
+#'
+#' @export
+#' @examples
+#' check_equal(1:2, 1:2) # TRUE
+#' check_equal(1:2, 2:3) # "Not all equal"
+#' check_equal(Inf, Inf) # "Not all equal"
+#' check_equal(0.01, 0.02) # "Not all equal"
+#' check_equal(0.01, 0.02, tol = 0.05) # TRUE
+#' check_equal(1, c(1, 1)) # "Not all equal"
```
danielinteractive, line 55: please remove comments - that is untypical for examples in R

R/checkmate.R
```diff
@@ -21,6 +21,107 @@
...
+check_equal <- function(..., tol = sqrt(.Machine$double.eps)) {
+  dot_args <- list(...)
+
+  sapply(dot_args, assert_numeric)
```
danielinteractive, line 59: better use `vapply` everywhere here (because it is safer)

R/checkmate.R
```diff
@@ -21,6 +21,107 @@
...
+    return("Some entries NA")
+  }
+  if (any(sapply(dot_args, is.infinite))) {
+    return("Not all entries finite")
+  }
+  if (!all(sapply(dot_args, test_numeric))) {
+    return("Not all numeric")
+  }
+
+  rv <- test_true(
```
danielinteractive, line 75: please replace `rv` by more meaningful variable name

R/checkmate.R
```diff
@@ -21,6 +21,107 @@
...
+      sapply(
+        2:length(dot_args),
+        function(z) abs(dot_args[[1]] - dot_args[[z]]) < tol
+      )
+    )
+  )
+  if (rv) {
+    return(TRUE)
+  } else {
+    return("Not all equal")
```
danielinteractive, line 86: avoid explicit return here at the end

R/checkmate.R
```diff
@@ -21,6 +21,107 @@
...
+  }
+}
+
+#  assert equality
```
danielinteractive, line 90:
```suggestion
```

R/checkmate.R
```diff
@@ -21,6 +21,107 @@
...
+#' @return `list(...)`, invisibly.
+#' @seealso [`assertions`] for more details.
+#' @inheritParams checkmate::assert_numeric
+#'
+#' @export
+#' @examples
+#' assert_equal(1:2, 1:2) # no error
+#' assert_equal(0.01, 0.02, tol = 0.05) # no error
+# nolint start
+assert_equal <- function(..., tol = sqrt(.Machine$double.eps), .var.name = vname(x), add = NULL) {
```
danielinteractive, line 119: better use `makeAssertionFunction` or add comment why that would not work

R/checkmate.R
```diff
@@ -21,6 +21,107 @@
...
+  if (rv) {
+    return(TRUE)
+  } else {
+    return("Not all equal")
+  }
+}
+
+#  assert equality
+
+#' Assert That All Arguments Are Equal
```
danielinteractive, line 92: please document in same page using `@rdname`, compare other extensions below

tests/testthat/test-checkmate.R
```diff
@@ -1,3 +1,25 @@
...
+  expect_true(check_equal(1:2, 1:2))
+  expect_equal(check_equal(1:2, 2:3), "Not all equal")
+  expect_equal(check_equal(Inf, Inf), "Not all entries finite")
+  expect_equal(check_equal(NA, 1), "Some entries NA")
+  expect_equal(check_equal(0.01, 0.02), "Not all equal")
+  expect_true(check_equal(0.01, 0.02, tol = 0.05))
+  expect_equal(check_equal(1, c(1, 1)), "Not all of same length")
+})
+
+# assert_equal
```
danielinteractive, line 12:
```suggestion
# assert_equal ----
```

danielinteractive (commented):
Thanks @Puzzled-Face , nice, please see comments below

R/checkmate.R
Puzzled-Face: > better use `vapply` everywhere here (because it is safer)

`vapply` is new to me. Not sure I agree here.  

The potential problem is `vapply`'s `FUN.VALUE` argument.  From the online doc: "all values of FUN [should be] compatible with the FUN.VALUE, in that they must have the same length and type".  

The problem is that we don't know:

1. How many entries are in `...`
2. The length of each entry in `...`

and, in addition,

3. `assert_numeric` returns an object of the same type as its first argument (or throws an error)

I *think* we could work all that out, but it would add considerably to the complexity of the function.

R/checkmate.R
Puzzled-Face: > better use `makeAssertionFunction` or add comment why that would not work

`assert_equal <- makeAssertionFunction(check_equal)` fails with error "Error in `checkmate::makeAssertion(..., res, .var.name, add)`: unused argument (add)".  I *think* this is because of the use of `...` in `check_equal`.

danielinteractive (approved):

PR #713, opened 2023-10-18
659 add default no argument constructors for remaining classes

Puzzled-Face:
Default constructors now exist for all crmPack classes.  For the following classes, the default constructor throws an error indicating that the class should not be instantiatied.

- `CohortSize`
- `CrmPackClass`
- `DualEndpoint`
- `GeneralData`
- `GeneralModel`
- `GeneralSimulationsSummary`
- `Increments`
- `ModelEff`
- `ModelPseudo`
- `ModelTox`
- `NextBest`
- `positive_number`
- `PseudoSimulations`
- `PseudoDualSimulations`
- `PseudoDualSimulationsSummary`
- `PseudoDualFlexiSimulations`
- `PseudoFlexiSimulations`
- `PseudoSimulationsSummary`
- `SimulationsSummary`
- `Report`
- `SafetyWindow`
- `Stopping`
- `Validate`

I am confident that the majority of these are virtual common ancestors of concrete subclasses and so throwing the error is appropriate.  However,  for some, marked with an asterisk, I would like confirmation.  In particular, I am puzzled by the the various `XxxxSimulations` and `XxxxSimulationsSummary` classes.  Many of them do not have constructors or example files, just class definitions.  For some `XxxxSimulationsSummary` classes, the example file simply wraps an instance of the corresponding `XxxxSimulations` class in a call to `summary`.  See, for example, `examples/Simulations-method-plot-DualSimulatuionsSummary.R`.


tests/testthat/test-CrmPackClass-class.R
```diff
@@ -30,4 +36,20 @@ test_that("CrmPackClass does not identify random non-crmPack classes", {
   }
 })
 
+test_that("virtual CrmPackClass classes throw expcted error when default constructor called", {
```
danielinteractive, line 39:
```suggestion
test_that("virtual CrmPackClass classes throw expected error when default constructor called", {
```

danielinteractive (approved):
Thanks @Puzzled-Face !
Yeah for the special designs we have special simulation and therefore also simulation summary classes, but it should be good like this

PR #718, opened 2023-10-26
716 plot methods do not always work as expected

### LINKED ISSUE #716: `plot` methods do not always work as expected
Take, for example, the code in `examples/Data-method-plot-DataDual.R`:

```
my_data <- DataDual(
  w = rnorm(8),
  doseGrid = c(0.1, 0.5, 1.5, 3, 6, seq(from = 10, to = 80, by = 2))
)

plot(my_data)
```

This code produces the following output

```
gtable[arrange]
```

rather than the expected plot(s).

The problem is caused by  calls to `gridExtra::arrangeGrob`, the output from which needs to be wrapped in a call to `grid::grid.draw` for the behaviour to be as expected.   For example, amending the final line of the snippet above to

```
grid.draw(plot(my_data))
```

yields

image

I do not know when this problem first emerged, nor why existing tests are currently not failing.

Calls to `arrangeGrob` appear in

- `R/Data-methods.R`
- `R/Rules-methods.R`
- `R/Samples-methods.R`
- `R/Simulations-methods.R`
- `R/helpers_rules.R`

In amending the existing code, care may need to be taken when handling code that involves `callNextMethod`.

Puzzled-Face:
Fixes #716 .


R/Data-class.R
```diff
@@ -5,7 +5,7 @@ NULL
 
 # GeneralData-class ----
 
-#' `GeneralData`
+#' `GeneralData` A Class for General Data Input
```
danielinteractive, line 8: I think we also don't have more in the title for the other data classes, so don't need that here then either?

danielinteractive (commented):
Thanks @Puzzled-Face for working on this, my concern with the `grid::grid.draw()` wrapper applied everywhere is that 1) this will no longer return the plot objects (because the value of it is NULL) and 2) it will always plot things on the device, where the user might just want to save plots first and look at them later.

Puzzled-Face:
> Thanks @Puzzled-Face for working on this, my concern with the `grid::grid.draw()` wrapper applied everywhere is that 1) this will no longer return the plot objects (because the value of it is NULL) and 2) it will always plot things on the device, where the user might just want to save plots first and look at them later.

Hmm.  Yes.  Good point.  (Though `grid.draw` is used in several places in the package already...) An alternative then would be to replace with calls to (eg) `gridExtra::grid.arrange()`, `ggpubr::ggarrange()` or `cowplot::plot_grid()`. `gridExtra::grid.arrange()` appears to be the best option as the other two would introduce a new dependency.  Or, less appealing (IMHO), would be to simply return a `list` and expect the end-user to sort things out for themselves - which would be a breaking change.

Should I also remove all other calls to `grid.draw`, or are they appropriate elsewhere?  (I am suspicious of graphs which always appear on an interactive device...)

danielinteractive:
@Puzzled-Face the problem is that also `grid.arrange` directly plots the object. My wish with the print and plot methods was to decouple the actual computation of the plot from the plotting (as it works e.g. with standard `ggplot2` objects)

Puzzled-Face:
> @Puzzled-Face the problem is that also `grid.arrange` directly plots the object. My wish with the print and plot methods was to decouple the actual computation of the plot from the plotting (as it works e.g. with standard `ggplot2` objects)

I should read the online doc more carefully!

We can wrap `ggpubr::grid.arrange()` in `ggpubr::as_ggplot()`.  The documentation for `cowplot::plot_grid()` is incomplete, but the function does appear to return a plot object:

```
library(cowplot)
library(tidyverse)

set.seed(123)
d <- lapply(1:3, \(i) tibble(x = rnorm(5), y = rnorm(5)))
p <- lapply(d, \(df) df %>% ggplot() + geom_point(aes(x =x, y = y)))
g <- plot_grid(plotlist = p)
class(g)
[1] "gg"     "ggplot"
```

`ggpubr::ggarrange()` returns a `ggplot` or a list of `ggplot`s.

```
library(ggpubr)
library(tidyverse)

set.seed(123)
d <- lapply(1:3, \(i) tibble(x = rnorm(5), y = rnorm(5)))
p <- lapply(d, \(df) df %>% ggplot() + geom_point(aes(x =x, y = y)))

g <- ggarrange(plotlist = p)
class(g)
[1] "gg"        "ggplot"    "ggarrange"
```

Reverting #697 remains an option.

How would you like to proceed?

danielinteractive:
Thanks @Puzzled-Face , to me reverting #697 still seems easiest but also most clean from a design perspective (basically we just want that the `gtable` object behaves as a `ggplot2` object and plots when we print it)

PR #720, opened 2023-10-26
719: add parallel_start

### LINKED ISSUE #719: Add a new option to allow parallel start of mono and combo for `DesignGrouped`
Use case:
- we have the first cohort only mono
- afterwards second cohort mono and combo in parallel
- then we might want to have the same dose in the second cohort
- from there on allow different doses in principle

To do:
- [ ] Add new slot to the `DesignGrouped` class to this effect

danielinteractive:
closes #719


R/Design-class.R
```diff
@@ -887,6 +887,11 @@ DADesign <- function(model, data,
 #'   concurrent mono and combo cohorts.
 #' @slot same_dose (`flag`)\cr whether the lower dose of the separately determined mono and combo
 #'   doses should be used as the next dose for both mono and combo.
+#' @slot parallel_start (`flag`)\cr whether for the time when mono and combo are the first time together,
+#'   the same dose should be used. Note that this is different from `same_dose` which will always force
+#'   them to be the same. This is useful as an option when we want to use `first_cohort_mono_only = TRUE`
```
Puzzled-Face, line 892: The description uses slightly awkward phrasing.  Perhaps "indicates whether, when mono and combo cohorts run in parallel for the first time, the same dose should be used. ..."

examples/Design-class-DesignGrouped.R
```diff
@@ -57,13 +57,17 @@ design <- DesignGrouped(
...
 
-# Alternative options: Here e.g. use both mono and combo from the start,
-# but allow different dose levels for the cohorts. Stop mono arm too, when combo
-# arm is stopped.
+# Alternative options: Here e.g.
+# - use both mono in first cohort and afterwards have mono and combo in parallel,
+# - but allow different dose levels for the cohorts.
+# - Stop mono arm too, when combo arm is stopped.
+# - Start parallel though, i.e. for the second cohort where mono and combo start together,
+#   use the same dose
```
Puzzled-Face, line 65: Again, the language here is awkward. So awkward, I'm not sure what you are trying to say!  "start in parallel" suggests that both mono and combo are used in the first cohort, so "for the second cohort where mono and combo start together..." is contradictory.

Puzzled-Face (changes requested):
See suggestions below, all related to comments.

danielinteractive:
Hi @Puzzled-Face when you get the chance could you re-review this? thanks :-)

R/Design-class.R
```diff
@@ -885,15 +885,26 @@ DADesign <- function(model, data,
 #' @slot first_cohort_mono_only (`flag`)\cr whether first test one mono agent cohort, and then
 #'   once its DLT data has been collected, we proceed from the second cohort onwards with
 #'   concurrent mono and combo cohorts.
-#' @slot same_dose (`flag`)\cr whether the lower dose of the separately determined mono and combo
+#' @slot same_dose_for_all (`flag`)\cr whether the lower dose of the separately determined mono and combo
 #'   doses should be used as the next dose for both mono and combo.
```
Puzzled-Face, line 889: Would "should be used as the next dose for both mono and combo _in all cohorts_" be clearer?

R/Design-class.R
```diff
@@ -885,15 +885,26 @@ DADesign <- function(model, data,
...
+#' @slot same_dose_for_start (`flag`)\cr indicates whether, when mono and combo are
+#'   used in the same cohort for the first time, the same dose should be used for both.
+#'   Note that this is different from `same_dose_for_all` which will always force
+#'   them to be the same. If `same_dose_for_all = TRUE`, this is therefore ignored. See details.
```
Puzzled-Face, line 893: The tag is `@details`, but the text seen by the reader is "Details".  Therefore, "See Details".

tests/testthat/test-Design-validity.R
```diff
@@ -72,14 +72,16 @@ test_that("v_design_grouped passes for valid object", {
 test_that("v_design_grouped messages wrong flag slots as expected", {
```
Puzzled-Face, line 72: "messages" could be either a verb or a noun.  Would "identifies" be clearer?

R/Design-class.R
danielinteractive: yes

Puzzled-Face (approved):

PR #722, opened 2023-10-27
Reverting #697

Puzzled-Face:


R/helpers.R
```diff
@@ -219,6 +219,24 @@ crmPackHelp <- function() {
   utils::help(package = "crmPack", help_type = "html")
 }
 
+## this is the new version, working on the gtable objects:
```
danielinteractive, line 222:
```suggestion
```

danielinteractive (commented):
Thanks @Puzzled-Face , it would be great to have one test where we see that an output from a plot method that returns a `gtable` is actually plotted when printed (so that we don't have a code regression risk later)

danielinteractive (approved):

PR #725, opened 2023-10-31
check that the dose limit is above the minimum of the dose grid, if t…

### LINKED ISSUE #482: nextBest: doselimit=0 leads to error
It is possible to specify a doselimit of 0 for the `nextBest` method. This leads to an error. Please see the following example.
Some checks may be necessary.

```
# Example of usage for `NextBestMTD` NextBest class.

# Create the data.
my_data <- Data(
  ID = 1:8,
  cohort = c(0, 1, 2, 3, 4, 5, 5, 5),
  doseGrid = c(0.1, 0.5, 1.5, 3, 6, seq(from = 10, to = 80, by = 2))
)

# Initialize the CRM model used to model the data.
my_model <- LogisticLogNormal(
  mean = c(-0.85, 1),
  cov = matrix(c(1, -0.5, -0.5, 1), nrow = 2),
  ref_dose = 56
)

# Set-up some MCMC parameters and generate samples from the posterior.
my_options <- McmcOptions(burnin = 100, step = 2, samples = 500)
my_samples <- mcmc(my_data, my_model, my_options)

# Define the rule for dose increments and calculate the maximum dose allowed.
my_increments <- IncrementsRelative(
  intervals = c(0, 20),
  increments = c(1, 0.33)
)
next_max_dose <- maxDose(my_increments, data = my_data)

# Define the rule which will be used to select the next best dose
# based on the 'NextBestMTD' class.
mtd_next_best <- NextBestMTD(
  target = 0.33,
  derive = function(mtd_samples) {
    quantile(mtd_samples, probs = 0.25)
  }
)

# Calculate the next best dose.
dose_recommendation <- nextBest(
  nextBest = mtd_next_best,
  doselimit = 0,
  samples = my_samples,
  model = my_model,
  data = my_data
)

dose_recommendation
```

```
$value
numeric(0)

$plot
Error in `check_aesthetics()`:
! Aesthetics must be either length 1 or the same as the data (1): x
Run `rlang::last_error()` to see where the error occurred.
```

dimitriskontosBAY:
# Pull Request

Fixes #482


danielinteractive (commented):
Thanks @dimitriskontosBAY , I would recommend:
1) let's add a test in any case that checks this 
2) if that does not lead to the problem reported by #482 then this is all we need
3) if that leads to a problem then let's modify the code in the generic, similar as you have now (but we only need this if we see a problem, see step 2)

dimitriskontosBAY:
Thanks @danielinteractive.
I have now added a test ('nextBest-NextBestMTD returns correct next dose and plot when doselimit=0') now for a start that is a pass. 
Could you please check if this is the right test we are after?

danielinteractive (approved):
Thanks @dimitriskontosBAY I think that is ok

PR #735, opened 2023-11-09
707 implement the mcmc prob dose fit methods for the logisticlognormalordinal class

Puzzled-Face:


R/Model-class.R
```diff
@@ -3335,3 +3335,98 @@ FractionalCRM <- function(...) {
...
+#' using an ordinal toxicity scale.
+#'
+#' @aliases LogisticLogNormalOrdinal
+#' @export
+.LogisticLogNormalOrdinal <- setClass(
+  Class = "LogisticLogNormalOrdinal",
+  contains = "ModelLogNormal",
+  slots = c(datablock = "function")
+  # ,
+  # validity = v_model_ordinal_logistic_log_normal
```
danielinteractive, line 3355: Would be good to have the validation function?

R/Model-class.R
```diff
@@ -3335,3 +3335,98 @@ FractionalCRM <- function(...) {
...
+#' @export
+.DefaultLogisticLogNormalOrdinal <- function() {
+  LogisticLogNormalOrdinal(
+    mean = c(-3, -4, 1),
+    cov = diag(c(3, 4, 1)),
+    ref_dose = 50
+  )
+}
+
+ordinal_data <- DataOrdinal(
```
danielinteractive, line 3425: This should not be here, I guess needs to go to example file

R/Samples-methods.R
```diff
@@ -921,6 +921,71 @@ setMethod("fit",
       return(ret)
     }
 )
+
+##' @param points at which dose levels is the fit requested? default is the dose
```
danielinteractive, line 925:
```suggestion
#' @param points at which dose levels is the fit requested? default is the dose
```
Etc below

R/Samples-methods.R
```diff
@@ -921,6 +921,71 @@ setMethod("fit",
...
+##' grid
+##' @param quantiles the quantiles to be calculated (default: 0.025 and
```
danielinteractive, line 927: Please don't repeat generic function argument description

danielinteractive (commented):
Thanks @Puzzled-Face looks nice, please see below comments

R/Model-class.R
Puzzled-Face: The `datablock` slot is a left over from a previous iteration of the design.  It's not necessary and I have deleted it.  Therefore, I think there is no need for a specific validation function for `LogisticLogNormalOrdinal`.  Do you agree?

R/Model-class.R
Puzzled-Face: Yep.  My bad.

R/Samples-methods.R
Puzzled-Face: Already fixed.

R/Model-class.R
danielinteractive: yep sounds good - then just delete the commented code

danielinteractive (approved):
Thanks @Puzzled-Face !

PR #737, opened 2023-11-13
736 implement rules methods for ordinal crm models

Puzzled-Face:
Here are first draft implementations of wrapper functions for `CohortSize`, `Increments`, `NextBest` and `Stopping` rules to handle  ordinal models.

Still to do:

- Implement tests
- Check that all variations in method parameters have been correctly handled

I would appreciate comments on the work I've done so far and confirmation that I'm heading in the right direction.


R/Rules-class.R
```diff
@@ -1434,6 +1489,58 @@ IncrementsMin <- function(increments_list) {
   )
 }
 
+# JK 13-Nov-2023 Start
```
danielinteractive, line 1492: just remember to remove these later :)

R/helpers_data.R
```diff
@@ -207,3 +207,34 @@ h_obtain_dose_grid_range <- function(object, ignore_placebo) {
...
+#' A simple helper function that takes a [`DataOrdinal`] object and an
+#' integer grade and converts them to the equivalent `Data` object.
+#'
+#' @param d (`DataOrdinal`)\cr the `DataOrdinal` object to covert
+#' @param grade (`integer`)\cr the toxicity grade for which the equivalent data
+#' is required.
+#' @return A [`Data`] object.
+#'
+#' @export
+h_convert_ordinal_data <- function(d, grade) {
```
danielinteractive, line 225: generally one letter argument names are not ideal (if not super standard like `x` etc.), how about `data`?

danielinteractive (commented):
Thanks @Puzzled-Face , looks great! Nice work

Puzzled-Face:
Tests now implemented.

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: crmPack
 Title: Object-Oriented Implementation of CRM Designs
-Version: 1.0.9000.9134
```
danielinteractive, line 3: yeah I would not bump this up manually, should be done automatically later by GH actions

R/Rules-class.R
```diff
@@ -1551,7 +1549,8 @@ IncrementsOrdinal <- function(grade, rule) {
 #'
 #' [`Stopping`] is a class for stopping rules.
 #'
-#' @slot report_label (`string`)\cr a label for the stopping report. The meaning
+#' @slot
```
danielinteractive, line 1552: seems the slot name is missing here

R/helpers_data.R
Puzzled-Face: Fair point.  I dislike `data` because of potential confusion with the function.  Shall we compromise on `df`?

DESCRIPTION
Puzzled-Face: I posted in Slack about this earlier today.  I have already reverted by hand.

R/Rules-class.R
Puzzled-Face: Hmmm.  Not one of my changes, but fixed.

R/helpers_data.R
danielinteractive: yeah sounds good

PR #745, opened 2023-11-22
doseFunction and probFunction for LogisticlogNormalOrdinal models

Puzzled-Face:


R/Model-methods.R
```diff
@@ -157,6 +206,56 @@ setMethod(
   }
 )
 
+## LogisticLogNormalOrdinal ----
+
+#' @describeIn doseFunction
```
danielinteractive, line 211: should this be 
```suggestion
#' @describeIn probFunction
```
?

R/mcmc.R
```diff
@@ -766,3 +766,24 @@ setMethod("mcmc",
     }
 )
 # nolint end
+
+## -----------------------------------------------------------------------------------
+## obtain the posterior samples for ordinal models
+## ----------------------------------------------------------------------------
```
danielinteractive, line 772: please avoid these delimiters - you can use `ordinal ----` instead e.g. (we might have some old delimiters in this file but please ignore those)

R/mcmc.R
```diff
@@ -766,3 +766,24 @@ setMethod("mcmc",
...
+setMethod(
+  f = "mcmc",
+  signature = signature(
+    data = "DataOrdinal",
+    model = "LogisticLogNormalOrdinal",
+    options = "McmcOptions"
+  ),
+  definition = function(data, model, options, ...) {
+    return_value <- callNextMethod()
+    names(return_value@data) <- gsub("\\[(\\d+)\\]", "\\1", names(return_value@data))
```
danielinteractive, line 786: Can you please add a comment here, and also a documentation sentence after `@describeIn` why this is needed / what it is doing?

R/Model-methods.R
Puzzled-Face: Yep.

R/mcmc.R
```diff
@@ -773,6 +773,14 @@ setMethod("mcmc",
 ##
 ##' @describeIn mcmc Obtain the posterior samples for the model parameters in the
 ##' `LogisticLogNormalOrdinal`.
+##'
+##' The generic `mcmc` method returns a `Samples` object with elements of the
```
danielinteractive, line 777:
```suggestion
##' For `LogisticLogNormalOrdinal`, the generic `mcmc` method returns a `Samples` object with elements of the
```
so that when reading the documentation page it is clear what this refers to

danielinteractive (approved):
Thanks @Puzzled-Face !

PR #746, opened 2023-11-22
723 remove unnecessary startup message

Puzzled-Face:
This should remove the "Creating a generic function for ‘get’ from package ‘base’ in package ‘crmPack’" message when `crmPack` is loaded.


R/Samples-methods.R
```diff
@@ -50,6 +50,8 @@ setMethod(
 ## plots with "ggmcmc" package
 ## --------------------------------------------------
```
danielinteractive, line 52: please add a comment to explain the necessity of this, otherwise we might remove this in the future.

danielinteractive (approved):

Issue #650, opened 2023-07-21
Ordinal CRM: Part 1


Puzzled-Face:
Begin implementation of the ordinal CRM:

- [x] The `DataOrdinal` class, including `update`, `dose_grid_range` and `plot` methods #666 
- [x] The `LogisticLogNormalOrdinal` class
- [x] `mcmc`, `prob`, `dose`,  `fit` methods
- [x] `plot-Samples-DataOrdinal` method
- [x] `doseFunction` and `probFunction`
- [x] First draft of a "How to" vignette
- [x] examples for all new methods and classes
- [x] tests for all new methods and classes

See #324 for design details

PR #752, opened 2023-11-24
747 vignette for ordinal crm

### LINKED ISSUE #747: Vignette for ordinal CRM
Write a vignette describing the use of the ordinal CRM classes.

See #650 and #324 for more information.

Puzzled-Face:
Closes #747 and #650.

### DIFF
--- a/R/Rules-methods.R
+++ b/R/Rules-methods.R
@@ -1679,6 +1679,24 @@ setMethod(
   }
 )
 
+#' @describeIn maxDose determine the maximum possible next dose based on
+#'   multiple increment rules, taking the minimum across individual increments.
+#'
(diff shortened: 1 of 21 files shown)


danielinteractive (approved):
Thanks @Puzzled-Face I understand that the vignette would be added in here but don't see it in the changed files? Otherwise looks good

Puzzled-Face:
> Thanks @Puzzled-Face I understand that the vignette would be added in here but don't see it in the changed files? Otherwise looks good

In my first push, I failed to rebuild `README` and update `_pkgdown.yaml`.  I fixed those and pushed again almost immediately.  Does that explain your comment?

As I look now (a day later), the link to the vignette appears in `README`, but generates a 404 error.  I will investigate.

PR #760, opened 2023-12-01
add test for edge case

### LINKED ISSUE #517: Add edge case tests for `maxDose-IncrementsRelativeParts`
To do:
- [ ] Add additional edge case tests for `maxDose` method for `IncrementsRelativeParts`, motivated by the question below
- [ ] Adapt any code if needed

History:

              > > Hi @danielinteractive - I think that here:
> > https://github.com/Roche/crmPack/blob/f041ae22d3991892257eaea30840ddbc804c73ac/R/Rules-methods.R#L1187
> > 
> > we need to extract the case when:
> > ```
> > 
> > (lastDoseLevel + increments@dlt_start) >= length(data@part1Ladder)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think in such a case we should call `callNextMethod(increments, data, ...)`, i.e. a usual relative increment rules apply.
> > What do you think?
> 
> Thanks Wojciech, I think we cannot refer to the parent method because we still need to use the ladder slot to select the dose?

yap, but how can we use it since we are already exceeding `part1Ladder` "scale", and also we are already in part2....

_Originally posted by @wwojciech in https://github.com/Roche/crmPack/issues/515#issuecomment-1491485824_

pengguanya:
# Pull Request

Fixes #517

add test for
https://github.com/Roche/crmPack/blob/bf87e968e29a6f95358122914ddc237107e84603/R/Rules-methods.R#L1567

### DIFF
--- a/tests/testthat/test-Rules-methods.R
+++ b/tests/testthat/test-Rules-methods.R
@@ -1307,7 +1307,8 @@ test_that("maxDose-IncrementsRelativeParts throws error when part1Ladder is exce
   )
   expect_error(
     maxDose(increments, data),
-    "Assertion on 'new_max_dose_level <= length(data@part1Ladder)' failed: Must be TRUE.", fixed = TRUE
+    "Assertion on 'new_max_dose_level <= length(data@part1Ladder)' failed: Must be TRUE.",
+    fixed = TRUE
   )
 })
 


danielinteractive (commented):
Thanks @pengguanya , but does this cover the case where `(lastDoseLevel + increments@dlt_start) >= length(data@part1Ladder)` so that we get a problem with https://github.com/Roche/crmPack/blob/f041ae22d3991892257eaea30840ddbc804c73ac/R/Rules-methods.R#L1187?

pengguanya:
> Thanks @pengguanya , but does this cover the case where `(lastDoseLevel + increments@dlt_start) >= length(data@part1Ladder)` so that we get a problem with
> 
> https://github.com/Roche/crmPack/blob/f041ae22d3991892257eaea30840ddbc804c73ac/R/Rules-methods.R#L1187
> 
> ?

The IncrementsRelativeParts has been changed already in the main branch. An assertion has been added there: https://github.com/Roche/crmPack/blob/bf87e968e29a6f95358122914ddc237107e84603/R/Rules-methods.R#L1567

I am not sure the code you show in commit [f041ae2](https://github.com/Roche/crmPack/commit/f041ae22d3991892257eaea30840ddbc804c73ac) is still valid.

PR #761, opened 2023-12-01
Fix #759

### LINKED ISSUE #759: `tidy-NextBestNCRM` produces incorrect results
```
x <- NextBestNCRM(
     target = c(0.2, 0.35),
    overdose = c(0.75, 1.00),
    max_overdose_prob = 0.1
)

x |> tidy()
```

Produces no visible output.  But

```
class(x |> tidy())
[1] "tbl_NextBestNCRM" "tbl_df"           "tbl"              "data.frame" 
```
as expected.

  - [ ] Update tests to catch the error
  - [ ] Correct the method

Puzzled-Face:
closes #759

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: crmPack
 Title: Object-Oriented Implementation of CRM Designs
-Version: 1.0.9000.9137
+Version: 1.0.9000.9138
 Authors@R: c(
     person("Daniel", "Sabanes Bove", , "daniel.sabanesbove@gmx.net", role = c("aut", "cre")),
     person("Wai", "Yin Yeung", , "winnie.yeung@roche.com", role = "aut"),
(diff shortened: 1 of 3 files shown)


tests/testthat/test-Rules-methods.R
```diff
@@ -4361,10 +4361,18 @@ test_that("tidy-IncrementsRelativeParts works correctly", {
   expect_snapshot_value(result, style = "deparse")
 })
 
+# Relevant:https://github.com/Roche/crmPack/issues/759
 test_that("tidy-NextBestNCRM works correctly", {
   obj <- .DefaultNextBestNCRM()
   result <- tidy(obj)
-  expect_snapshot_value(result, style = "deparse")
+  expected <- tibble::tibble(
```
danielinteractive, line 4368: ah yeah, so this is a good example where snapshot testing deceived us into thinking everything is fine

danielinteractive (approved):

tests/testthat/test-Rules-methods.R
Puzzled-Face: ... but also partly my fault for not checking the original snapshot to notice that its value was incorrect.  Subsequent snapshots passed because the incorrect value didn't change.  Human eyeball Mk 1 picked up the problem as soon as it appeared in real life.

Issue #682, opened 2023-09-22
Clean up `fromQuantiles.R`


danielinteractive:
To do:

Clean up in below steps by separate PRs (to avoid too large PRs).

When somebody takes a task please put your name next to it (edit this description) so we don't duplicate any work :) thanks!

- [x] `Quantiles2LogisticNormal` -> @0liver0815 
- [x] `getMinInfBeta` -> @danielinteractive 
- [x] `MinimalInformative`

PR #773, opened 2024-01-15
clean up function `getMinInfBeta` and add tests

danielinteractive:
part of #682


R/fromQuantiles.R
```diff
@@ -1,23 +1,8 @@
...
-## Find the best LogisticNormal model for a given set of quantiles at certain
-## dose levels
-##
-## History:
-## 11/02/2014   file creation
-#####################################################################################
-
 ##' @include helpers.R
 ##' @include Model-class.R
 {}
```
dimitriskontosBAY, line 3: @danielinteractive  this seems to trigger an error. 
I am not sure why this is used here like this and if putting {} in diffetrnt lines would make sense?

danielinteractive, line 3:
```suggestion
NULL
```

dimitriskontosBAY (approved):
thanks @danielinteractive no further comments

Issue #775, opened 2024-01-16
Implement `knit_print` for subclasses of `CohortSize` (including tests)


Puzzled-Face:
See #774 for details.

PR #776, opened 2024-01-16
771 change from base pipe | to magrittr pipe

### LINKED ISSUE #771: Change from base pipe `|>` to `magrittr` pipe `%>%`
... to be compatible with older R versions.

Then also run `R CMD check` on R 4.0.4 at least (that could be our minimum version required)

Puzzled-Face:
Fixes #771


Puzzled-Face:
`helpers_knit_CohortSize.R` belongs in #775.  I will delete.

PR #778, opened 2024-01-19
fix header

### LINKED ISSUE #777: `GeneralSimulationsSummary` @pengguanya

pengguanya:
# Pull Request

Fixes #777

### DIFF
--- a/R/Simulations-class.R
+++ b/R/Simulations-class.R
@@ -286,49 +286,47 @@ DualSimulations <- function(rho_est,
   )
 }
 
-# GeneralSimulationsSummary ----
-
-# nolint start
-##' Class for the summary of general simulations output
-##'
-##' Note that objects should not be created by users, therefore no
-##' initialization function is provided for this class.
-##'
-##' @slot target target toxicity interval
-##' @slot targetDoseInterval corresponding target dose interval
-##' @slot nsim number of simulations
-##' @slot propDLTs proportions of DLTs in the trials
-##' @slot meanToxRisk mean toxicity risks for the patients
-##' @slot doseSelected doses selected as MTD
-##' @slot toxAtDosesSelected true toxicity at doses selected
-##' @slot propAtTarget Proportion of trials selecting target MTD
-##' @slot doseMostSelected dose most often selected as MTD
-##' @slot obsToxRateAtDoseMostSelected observed toxicity rate at dose most often
-##' selected
-##' @slot nObs number of patients overall
-##' @slot nAboveTarget number of patients treated above target tox interval
-##' @slot doseGrid the dose grid that has been used
-##' @slot placebo set to TRUE (default is FALSE) for a design with placebo
-##'
-##' @export
-##' @keywords classes
+#' `GeneralSimulationsSummary`
+#'
+#' @description `r lifecycle::badge("stable")`
+#'
+#' This class captures the summary of general simulations output. Note that objects
+#' should not be created by users, therefore no initialization
+#' function is provided for this class.
+#'
+#' @slot target (`numeric`)\cr target toxicity interval
+#' @slot target_dose_interval (`numeric`)\cr corresponding target dose interval
+#' @slot nsim (`integer`)\cr number of simulations
+#' @slot prop_dlts (`ANY`)\cr proportions of DLTs in the trials
+#' @slot mean_tox_risk (`numeric`)\cr mean toxicity risks for the patients
+#' @slot dose_selected (`numeric`)\cr doses selected as MTD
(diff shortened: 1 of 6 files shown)


R/Simulations-class.R
```diff
@@ -286,49 +286,47 @@ DualSimulations <- function(rho_est,
...
+#' @description `r lifecycle::badge("stable")`
+#'
+#' This class captures the summary of general simulations output. Note that objects
+#' should not be created by users, therefore no initialization
+#' function is provided for this class.
+#'
+#' @slot target (`numeric`)\cr target toxicity interval
+#' @slot target_dose_interval (`numeric`)\cr corresponding target dose interval
+#' @slot nsim (`integer`)\cr number of simulations
+#' @slot prop_dlts (`ANY`)\cr proportions of DLTs in the trials
```
danielinteractive, line 300: are you sure any type works here? I would guess it should be `numeric` or something specific?

R/Simulations-class.R
```diff
@@ -286,49 +286,47 @@ DualSimulations <- function(rho_est,
...
+#' @slot mean_tox_risk (`numeric`)\cr mean toxicity risks for the patients
+#' @slot dose_selected (`numeric`)\cr doses selected as MTD
+#' @slot tox_at_doses_selected (`numeric`)\cr true toxicity at doses selected
+#' @slot prop_at_target (`numeric`)\cr Proportion of trials selecting target MTD
+#' @slot dose_most_selected (`numeric`)\cr dose most often selected as MTD
+#' @slot obs_tox_rate_at_dose_most_selected (`numeric`)\cr observed toxicity rate at dose most often selected
+#' @slot n_obs (`ANY`)\cr number of patients overall
```
danielinteractive, line 307: same question here

R/Simulations-class.R
danielinteractive: note that the class in the S4 definition below can be more generic (e.g. here `ANY`) but we might have/need extra validation check for these which further restricts it and this should be the type listed here in the user documentation.

tests/testthat/test-Simulations-class.R
```diff
@@ -181,3 +181,46 @@ test_that("DualSimulations user constructor arguments names are as expected", {
...
+# GeneralSimulationsSummary ----
+test_that("GeneralSimulationsSummary generates object correctly", {
+  target_value <- 1
+  target_dose_interval_value <- 2
+  nsim_value <- 3L
+  mean_tox_risk_value <- 4
+  dose_selected_value <- 5
+
+  result <- expect_silent(
+    new("GeneralSimulationsSummary",
```
danielinteractive, line 194: I would use here the internal constructor `.GeneralSimulationsSummary()` instead of `new`

R/Simulations-class.R
danielinteractive: ok in this case it is an abstract parent class. then I would not say anything here maybe about the type, or look at the child classes what they have in terms of type for these arguments

danielinteractive (commented):
Thanks @pengguanya ! nice work, see minor questions on types below

R/Simulations-class.R
pengguanya: @danielinteractive Thanks for the comment. Looking into the downstream code. They (`nObs` and `propPLTs`) could be list, but they could also be numeric arrays in multi-dimensions. Below are the examples where they are arrays. But at somewhere else they could also be a one deferential list. I found no constrains or checks on exact type or even just dimensions. Only thing a bit more certain is their contents should be numeric, because there are computational operations (*, +, mean...) applied on them. Should I just add description that they could be numeric array or list and then keep the "ANY"? 
```
      ## number of patients overall
      if (object@data[[1]]@placebo) {
        nObs <- sapply(
          object@data,
          function(x) {
            data.frame(
              n.ACTV = sum(x@xLevel != 1L),
              n.PLCB = sum(x@xLevel == 1L)
            )
          }
        )
        nObs <- matrix(unlist(nObs), dim(nObs))
      } else {
```
```
          plotList[[plotIndex <- plotIndex + 1L]] <-
            h_barplot_percentages(
              x = x@propDLTs[1, ] * 100,
              description = "Proportion of DLTs [%] on active",
              xaxisround = 1
            )
        }
```

danielinteractive: ok wow, thanks for digging into this and finding this. ok yeah let's do as you propose

R/Simulations-class.R
danielinteractive: can resolve similarly here, but I would hope that `n_obs` would always be a `count`?...

danielinteractive (approved):

Issue #774, opened 2024-01-16
Implement `knit_print` and `render` methods for reporting


Puzzled-Face:
See #768 for design details.

This is the issue to manage implementation of S3 `knit_print` and S4 `render` methods for `crmPack`.  Please use the bullet list below to manage the implementation of each sub task in the list.  Edit this issue to indicate that you have taken responsibility for each sub-task.

To avoid merge conflicts, I suggest adding new features in files named `helpers_knitr_<subtask>.R`.  Once all sub tasks have been completed, the final step can be to combine all the task-specific files into `helpers-knitr.R`.

The `knit_print` methods for each sub-class of `GeneralModel` will be the longest and most fiddly to write.  For this reason I have separated them out into individual sub-tasks.  It might be possible to combine the implementation of one or two methods into a single PR.

### To do
 
- [ ] Implement `knit_print` for `subclasses of `Data` (including tests)
- [ ] Implement `knit_print` for `subclasses of `CohortSize` (including tests)
- [ ] Implement `knit_print` for `subclasses of `Increments` (including tests)
- [ ] Implement `knit_print` for `subclasses of `NextBest` (including tests)
- [ ] Implement `knit_print` for `subclasses of `Stopping` (including tests)
- [ ] Implement `knit_print` for `subclasses of `CohortSize` (including tests)
- [ ] Implement `knit_print` for `subclasses of `Model` (including tests)
- [ ] Implement `knit_print` for `subclasses of `Design` (including tests)
- [ ] Implement `knit_print` for `subclasses of `Simulations` (including tests)
- [ ] Implement `knit_print` for `subclasses of `SimulationSummary` (including tests)
- [ ] Implement `render` (including tests)
- [ ] Ensure "check method exists" test fails when a `knit_print` method is not found rather than simply printing a warning
- [ ] Convert design document to vignette
- [ ] Combine task-specific files into a single file

Puzzled-Face:
It appears that the S3 `knit_print` methods for each class are not registered until they are documented with `devtools::document`.  Until then, they must be registered manually.  For example

```
registerS3method("knit_print", "IncrementsRelativeDLTCurrent", knit_print.IncrementsRelativeDLTCurrent)
```

Issue #734, opened 2023-11-09
Testing for method existence


Puzzled-Face:
The unit testing in `crmPack` currently focuses on the correct operation of methods and functions that exist.  It does not check that all functions that **_should_** exist actually **_do_** exist.  A recent example of this was the need to define some missing default constructors [Issue #659].

Similarly, we currently rely on good review to ensure that all required methods for a new class have been implemented.  [For example, `dose`, `prob` and `fit` methods for a new `Model`/`Data` combination.] I think it would be helpful to automate such checking.

Consider

```
#' Tests For Existence of Required Method for ALl Relevant Subclasses
#' 
#' Examines all subclasses of the parent superclass to determine if a method 
#' with the given name and signature exists.  '{cls}' is used as a placeholder
#' for the subclass name in the required signature
#' 
#' @param method_name ('character')\cr the name of the method
#' @param parent_class ('character')\cr the name of the parent_class
#' @param signature ('character')\cr vector defining the default signature of 
#' the method
#' @param exclude ('character')\cr a scalar or vector defining the subclasses for
#' no method is expected to exist and therefore the test should be skipped.  Use
#' only for virtual subclasses 
#' @param exceptions (`character`) a named list defining the class specific signatures
#' for which the default signature should not be used.  The names of the list define
#' the class name, the corresponding value the signature to be used.
#' @returns `TRUE` if all required methods exist, otherwise a character of the form
#' `method <xxx> not found for classes <yyy>, <zzz>`
#' @internal
h_all_required_methods_exist <- function(method_name, parent_class, signature, exclude = c(), exceptions = list()) {
  subclass_names <- names(getClass(parent_class)@subclasses)
  subclass_names <- setdiff(subclass_names, as.vector(exclude, "character"))
  methods_exist <- unlist(
    lapply(
      subclass_names,
      function(cls) {
        this_signature <- if(cls %in% names(exceptions)) {
          this_signature <- exceptions[[cls]]
        } else {
          this_signature <- signature
        } 
        this_signature <- ifelse(this_signature == "{cls}", cls, this_signature)
        hasMethod(method_name, this_signature)
      }
    )
  )
  names(methods_exist) <- subclass_names
  if (all(methods_exist)) {
    return(TRUE)
  } 
  missing_classes <- which(!methods_exist)
  paste0(
    "method \"", 
    method_name, 
    "\" not found for class(es) ", 
    paste0(names(missing_classes), collapse = ", ")
  )
}
```

so that

```
[...]

PR #784, opened 2024-02-01
775 implement knit print for subclasses of cohortsize including tests

Puzzled-Face:
Implements #775, part 1 of #774.


Puzzled-Face:
Test coverage report suggests that `knit_print` methods for `CohortSizeMax`, `CohortSizeMin` and `CohortSizeOrdinal` are not covered.  But the contents of the `/_snaps/helpers_knitr` clear shows that they are.

R/helpers_knitr_CohortSize.R
```diff
@@ -0,0 +1,229 @@
...
+  if (length(label) == 1) {
+    label[2] <- paste0(label[1], "s")
+  }
+  rv <- paste0("A constant size of ", x@size, " ", label[ifelse(x@size == 1, 1, 2)], ".")
+  if (asis) {
+    rv <- knitr::asis_output(rv)
+  }
+  rv
+}
+registerS3method("knit_print", "CohortSizeConst", knit_print.CohortSizeConst)
```
danielinteractive, line 51: are you sure we need to do it like this? I was more expecting that the `@export` above is sufficient, since we now import `knitr` package and the generic function. See https://roxygen2.r-lib.org/articles/namespace.html#s3 for details

R/helpers_knitr_CohortSize.R
```diff
@@ -0,0 +1,229 @@
...
+  assert_flag(asis)
+
+  param <- list(...)
+  if (!("col.names" %in% names(param))) {
+    param[["col.names"]] <- c("Lower", "Upper", "Cohort size")
+  }
+  if (!("caption" %in% names(param))) {
+    param[["caption"]] <- "Defined by the dose to be used in the next cohort"
+  }
+  x <- x %>% tidy()
```
danielinteractive, line 70:
```suggestion
  x <- tidy(x)
```
generally let's avoid pipe in package code, and especially for such usage normal function call is shorter

R/helpers_knitr_CohortSize.R
```diff
@@ -0,0 +1,229 @@
...
+  assert_flag(asis)
+
+  param <- list(...)
+  if (!("col.names" %in% names(param))) {
+    param[["col.names"]] <- c("Lower", "Upper", "Cohort size")
+  }
+  if (!("caption" %in% names(param))) {
+    param[["caption"]] <- "Defined by the number of DLTs so far observed"
+  }
+  param[["x"]] <- x %>% tidy()
```
danielinteractive, line 104: similar here etc.

tests/testthat/test-helpers_knitr.R
```diff
@@ -0,0 +1,109 @@
...
+    "PseudoDualFlexiSimulations", "PseudoFlexiSimulations",
+    "PseudoSimulationsSummary", "SimulationsSummary", "Report", "SafetyWindow",
+    "Stopping", "Validate",
+    # The following classes have no constructors
+    "DualSimulationsSummary"
+  )
+  crmpack_class_list <- setdiff(crmpack_class_list, exclusions)
+
+  # See https://stackoverflow.com/questions/42738851/r-how-to-find-what-s3-method-will-be-called-on-an-object
+  identifyMethod <- function(generic, ...) {
```
danielinteractive, line 22: hmm........ this will be difficult to maintain. any way we can keep this much simpler for the tests?...

danielinteractive (commented):
Thanks @Puzzled-Face , nice work!
I have some questions here about the tests... can we maybe stick to a simpler, even though more lines of code involved, way of handling this?

R/helpers_knitr_CohortSize.R
Puzzled-Face: I was surprised it was necessary as well.  But it appeared to be necessary as I was developing.  But that may simply have been to fix the consequences of another mistake.  I will investigate.

tests/testthat/test-helpers_knitr.R
Puzzled-Face: I'm not sure which part you think will be difficult to maintain: the exclusion list, the code to "find" the S3 method for each class, or something else?  

The exclusion list is awkward, I agree.  But I think it's reasonably stable - it's essentially a list of the virtual classes that `crmPack` defines.  I don't see that changing very frequently.   I propose two ways of handling this in #734. [Note the paragraph that reads

> An alternative approach would be to require all methods to exist even for virtual superclasses. The actual method implementation for these virtual superclasses would then throw an exception (as is done by the default constructors for these super classes. This would simplify the unit testing, but at the cost of additional code that is never expected to be used.

I used the same approach to test that all the necessary `tidy` methods had been defined.

[As an aside, I think one weakness of our current tests is that we only test that what we have done works correctly.  We don't test that we have done everything we should have done. The approach I've taken here is, I think, the only way to address that weakness.]

If you mean that the code to find the appropriate S3 method is flaky, I disagree.  The fundamental process should hold true regardless of what happens in the future.  If not, it's likely that S3 is broken and we will have bigger issues to deal with!

If you mean the use of `is_checking()`, I think that can probably be removed now.  It was one of the things I tried to solve the problems I asked about in chat.  I think Craig spotted the root cause, so this is probably unnecessary now.

The use of `if (methodName != "knit_print.default")` is effectively temporary.  Once we have all the necessary `knit_print` methods implemented, we can either remove it entirely, or (my preference) replace it with `fail(paste0("Required knit_print method has not been defined for ", cls))` or similar.  There's a step to do this listed in #774.  Again, I took the same approach when testing the `tidy` methods.

Puzzled-Face: I can add more comments if you think that would be helpful.

danielinteractive: Basically the whole lines 22-63 are difficult to maintain. I don't understand what the code is doing to be honest. Plus there is a lot of typically dangerous things that work on the language, e.g. `deparse` , `substitute`, `assign`. 

My question is would it be possible to have much more vanilla kind of testing here, which is avoiding those dangerous things, avoiding loops where possible, and keep with a maybe tedious but explicit list of tests? 

Example: https://github.com/Roche/crmPack/blob/main/tests/testthat/test-Model-class.R I guess we could 
have tried to use loops to avoid manual writing of tests here too, but we kept with the much more simple structure.

If it is possible - let's go with the more simple approach. If not - then we need to derisk this code, by testing it explicitly (yes, testing the code that is used for tests here), documenting, commenting, etc.

Puzzled-Face: OK.  I hope I can reassure you.  The code that makes you uncomfortable is temporary.  Put simply, it says "do we have a custom `knit_print` method for the current class?  If so, test it.  If not, do nothing.".  So, once we (think we) have `knit_print` methods for every `crmPack` class, we can delete the conditional testing and just test everything.  

In this particular case, we could delete the conditional testing immediately.  That will mean that tests will fail when we introduce new `knit_print` methods, but I don't see that as a major problem: we'd have to inspect new snapshot files in that case anyway.

The more important reason for keeping the current structure is that I can re-use it in other situations.  For example, to test the behaviour of the `asis` parameter to `knit_print` for `crmPack` classes.  The generic `knir_print` doesn't have this, so we would need either to use a class-specific test (which I think is less reliable and definitely less efficient) or to use a similar approach.

I'll refactor to make the code more self-explanatory.

danielinteractive: Ok and let's try to also have tests for the test code, e.g. make a function for pieces of it, add documentation and tests. Otherwise somebody else cannot maintain this in the future

R/helpers_knitr_CohortSize.R
Puzzled-Face: Fixed.

tests/testthat/test-helpers_knitr.R
Puzzled-Face: Done. See revisions in the latest push

danielinteractive (approved):
nice work, thanks @Puzzled-Face !

PR #790, opened 2024-02-06
785 implement knit print for subclasses of increments

Puzzled-Face:
Part of #774.


R/helpers_kintr_Increments.R
```diff
@@ -0,0 +1,318 @@
+#' @importFrom knitr knit_print
```
danielinteractive, line 1: I think we should only have this statement once, so suggest to look for the best place and then have it there once

R/helpers_kintr_Increments.R
```diff
@@ -0,0 +1,318 @@
...
+NULL
+
+# Increments ----
+
+#' Render a IncrementsRelative Object
```
danielinteractive, line 6:
```suggestion
#' Render a `IncrementsRelative` Object
```
to avoid spelling problems as usual, same elsewhere

R/helpers_kintr_Increments.R
```diff
@@ -0,0 +1,318 @@
...
+#'
+#' @description `r lifecycle::badge("experimental")`
+#' @inherit knit_print.CohortSizeConst return
+#' @param ... passed to `knitr::kable`
```
danielinteractive, line 10:
```suggestion
#' @param ... passed to [knitr::kable()] 
```
so the reader can click on the link

tests/testthat/test-helpers_knitr.R
```diff
@@ -97,14 +100,20 @@ test_that("asis parameter works correctly for all implemented methods", {
...
+
+        # Explicit behaviours
         rv <- knit_print(obj, asis = TRUE)
         expect_class(rv, "knit_asis")
         rv <- knit_print(obj, asis = FALSE)
         # Most objects return a character, but not all.  For example,
         # CohortSizeDLT returns a knitr_table
         if ("knit_asis" %in% class(rv)) print(cls)
         expect_true(!("knit_asis" %in% class(rv)))
+
```
danielinteractive, line 114: are we sure we want to have all these html files in the snapshots? I have seen in the past that these can be very much depending on the system and package versions, so expect very frequent updates will be needed to avoid failing tests

danielinteractive (commented):
thanks @Puzzled-Face !

R/helpers_kintr_Increments.R
Puzzled-Face: Ultimately, when we merge all the class-specific `knit_print_<ClassName>.R` files into one, we will have only one instance.  I'll check to see if we can get away with only one in the mean time.

R/helpers_kintr_Increments.R
Puzzled-Face: Done.  Also for ``knit_print_CohortSize.R`, though it did not seem to be a problem.

tests/testthat/test-helpers_knitr.R
Puzzled-Face: As we've discussed previously, snapshots aren't the ideal solution, but they are better than nothing.  In this case, I think they are slightly more relevant than usual as the purpose of `knit_print` is to provide a user-friendly rendition of a `crmPack` object.  If the irritation of maintaining the snapshots outweighs the effort of creating class-specific tests, I will reassess my position! :-)

R/helpers_kintr_Increments.R
Puzzled-Face: Should this be [`knitr::kable`]?

R/helpers_kintr_Increments.R
danielinteractive: hm I would not merge into a huge file in the end, it will be too much at once. could also move one statement to the central package.R file

tests/testthat/test-helpers_knitr.R
danielinteractive: yeah I just remember what happened last time with snapshots - and html is even more fragile I think. if we can have your commitment to update the snapshots when it fails then ok

R/helpers_kintr_Increments.R
Puzzled-Face: Good idea.  Have placed it in `crmPack-package.R`.

R/helpers_kintr_Increments.R
Puzzled-Face: OK.

Puzzled-Face:
Please check the test results for `IncrementsRelativeParts` (in `/_snaps/helpers_knitr`) carefully.  I'm not absolutely certain I've understood what the class is attempting to do.

danielinteractive (approved):
Thanks @Puzzled-Face , just double checked the `html` output you mentioned and it looks right to me.

PR #800, opened 2024-03-06
781 cleanup pseudo simulations summary

### LINKED ISSUE #781: `PseudoSimulations` @pengguanya

pengguanya:
# Pull Request

Fixes #781


R/Simulations-class.R
```diff
@@ -496,119 +496,108 @@ DualSimulations <- function(rho_est,
...
-##' @export
+#' `PseudoSimulations`
+#'
+#' @description `r lifecycle::badge("stable")`
+#' This class captures trial simulations from designs using pseudo model.
+#' It has additional slots `fit` and `stop_reasons` compared to the
+#' general class [`GeneralSimulations`].
+#'
+#' @slot fit (`list`)\cr final fit values.
+#' @slot final_td_target_during_trial_estimates (`numeric`)\cr final estimates of the TDtargetDuringTrial.
```
danielinteractive, line 511:
```suggestion
#' @slot final_td_target_during_trial_estimates (`numeric`)\cr final estimates of the `TDtargetDuringTrial`.
```
please 1) use backticks around these code variable names, and 2) consider the need for making them snake_case, I don't see right away if they refer to old names or not

pengguanya: The slot is called "final_td_target_during_trial_estimates", not "td_target_during_trial" and there are many internal variables in different programs named as "TDtargetDuringTrial". I thought that could also mean something else. That is why I did not make the change in the description.

danielinteractive (approved):
Thanks @pengguanya , I think that is ok for now, since it is just a documentation update. Later when we would clean up e.g. https://github.com/Roche/crmPack/blob/main/R/Design-methods.R#L1913 then we can make the corresponding change to snake case too.

danielinteractive:
I think the vignette build is failing, can you have a look? Probably some downstream renaming needs to be done accordingly to the slot name updates.

PR #802, opened 2024-03-20
temporary suggestion for testing simulate for design class

### LINKED ISSUE #798: Code coverage for simulate for RuleDesign
Code clean sub taks from https://github.com/Roche/crmPack/issues/797

 simulate for RuleDesign (https://github.com/Roche/crmPack/blob/main/R/Design-methods.R#L514)

robertadamsbayer:
# Pull Request

Fixes #798


robertadamsbayer:
Hello @danielinteractive please review the suggestion for testing. I tried both, the scenario with fixed seeds (checking the full mySims object, thus inherently checking all subsequent scenarios) and the scenario which just checks for the "meta data" (i.e. lengths, classes etc.). Before continuing to apply that to all other methods in Design-methods I wanted to check in and ask for best-practice guidance. Also, I would suggest, if we agree on the testing strategy, to do all Design-methods testing in one branch / PR then because it will be quite straight forward to implement and review?

tests/testthat/test-helpers_design.R
```diff
@@ -205,3 +205,203 @@ test_that("h_determine_dlts returns correctly updated data object for placebo =
...
+  myTruth <- probFunction(model, alpha0 = 7, alpha1 = 8)
+
+  options <- McmcOptions(
+    burnin = 100,
+    step = 2,
+    samples = 1000,
+    rng_kind = "Mersenne-Twister",
+    rng_seed = 1234
+  )
+  time <- system.time(mySims <- simulate(design,
```
danielinteractive, line 269: do we need to time this simulation run? (not sure, because `time` is also not used below and would anyway differ between platforms)

tests/testthat/test-helpers_design.R
```diff
@@ -205,3 +205,203 @@ test_that("h_determine_dlts returns correctly updated data object for placebo =
...
+
+  expect_equal(mySims, expected)
+
+  expect_class(mySims, "Simulations")
+
+  expect_equal(any(sapply(mySims@fit[[1]], is.numeric)), TRUE) # check if all elements in mySims@fit are numeric
+
+  expect_equal(length(mySims@stop_report), 5) # check for length
+
+  expect_logical(mySims@stop_report) # check for stop_report to be logical vector
```
danielinteractive, line 406: something like this would be sufficient

tests/testthat/test-helpers_design.R
```diff
@@ -205,3 +205,203 @@ test_that("h_determine_dlts returns correctly updated data object for placebo =
...
+    mcmcOptions = options,
+    parallel = FALSE,
+    derive = list(
+      max_mtd = max,
+      mean_mtd = mean,
+      median_mtd = median
+    ),
+  ))[3]
+
+  expected <-
```
danielinteractive, line 283: this is too detailed / too much maintenance when numbers change etc.

danielinteractive (commented):
thanks @robertadamsbayer for the nice work! please see feedback below - I would simplify this further. yes, once agreed, we can put many tests in the same PR. no need to shoot for everything right away though, because then the PR gets again too large and takes too much time. incremental progress here is still best too

tests/testthat/test-helpers_design.R
robertadamsbayer: fully agree - was a bit lazy from me taking that over from the examples...

tests/testthat/test-helpers_design.R
robertadamsbayer: Thanks for guidance! I will adopt.

robertadamsbayer: @danielinteractive Hi Daniel, I put the new "end-to-end" test for simulate in "test-helpers_design.R". When moving it over to "test-Design-methods.R" I realized that there are also tests (using snapshots) for simulate already implemented. Those use fixed rng seeds, therefore "mimicking" quite closely what you suggested not to do. I don´t know what to keep and what to get rid of there?!

image

image

danielinteractive: I would just leave existing tests. More coverage does not hurt us

danielinteractive:
Thanks @robertadamsbayer , I would check the examples folder, if there is nothing it might be a bug ... and we would need to fix it

danielinteractive:
This PR is parked for now, in order to prioritize tests for Simulation methods.

danielinteractive (approved):

PR #812, opened 2024-04-18
807 implement knit print for subclasses of nextbest

Puzzled-Face:
Please check my interpretation of the methods for `NextBestTD`, `NextBestTSamples`, `NextBestMaxGain` and `NextBestMaxGainSamples` in particular.


R/helpers_knitr_NextBest.R
```diff
@@ -0,0 +1,621 @@
+# Integration with knitr ----
+#'
+#' @description `r lifecycle::badge("experimental")`
+#'
+#' We provide additional utility functions to allow human-friendly rendition of
+#' crmPack objects in Markdown and Quarto files.
+#'
+#' @return a character string that represents the object in markdown.
+#' @name knit_print
+NULL
```
danielinteractive, line 10: this should not be here/ only be in a single central place?
e.g. https://github.com/Roche/crmPack/blob/main/R/helpers_knitr_CohortSize.R already has this. maybe could be good to have this in a central `helpers_knitr.R` file.

R/helpers_knitr_NextBest.R
```diff
@@ -0,0 +1,621 @@
...
+#' @export
+#' @method knit_print NextBestDualEndpoint
+knit_print.NextBestDualEndpoint <- function(
+    x,
+    ...,
+    tox_label = "toxicity",
+    biomarker_label = "the biomarker",
+    biomarker_units = ifelse(x@target_relative, "%", ""),
+    asis = TRUE) {
+  # Validate
```
danielinteractive, line 180:
```suggestion
```
generally in here we avoid comments like this

R/helpers_knitr_NextBest.R
```diff
@@ -0,0 +1,621 @@
...
+  assert_flag(asis)
+  assert_character(tox_label, len = 1, any.missing = FALSE)
+  assert_character(biomarker_label, len = 1, any.missing = FALSE)
+  assert_character(biomarker_units, len = 1, any.missing = FALSE)
+
+  # Execute
```
danielinteractive, line 186:
```suggestion
```
similarly elsewhere

tests/testthat/test-helpers_knitr_NextBest.R
```diff
@@ -0,0 +1,503 @@
...
+      "Based on a toxicity grade of 2: The dose level recommended for the ",
+      "next cohort will be selected as follows:\n\n-  First, a custom ",
+      "measure of the posterior distribution of DLT will be calculated for all ",
+      "dose levels that are eligible according to the  Increments rule.\n",
+      "-  Next, the \"target dose\" (which may not be part of the dose grid) ",
+      "for which a custom measure of the posterior distribution of DLT is ",
+      "exactly equal to the target rate of 0.25 will be determined.\n",
+      "- Finally, the dose level whose absolute distance from the target dose ",
+      "is smallest will be selected as the recommended dose for the next cohort\n\n"
+    )
```
danielinteractive, line 501: these tests look good, thanks, but do we need all the new snapshots? at least in GitHub they are not useful because I just see the raw html code and not the page or contents. and I thought in the last PR we did not add any snapshots

R/helpers_knitr_NextBest.R
Puzzled-Face: We discussed whether the `knit_print` helpers should be in a single file or many previously.  Separate files was part of the original proposal (my preference), but you wanted to keep them separate.  Apologies, but I can't find the reference...

So far, the `knit_print` methods encompass 1518 statements, with methods for about another 10 classes/superclasses still to implement.  That would make a single file one of the largest in the package, so I've come round to your way of thinking.  It seems you've also come round to mine!

I don't think single file vs many files is a major issue.  we can do it either way.  If we stick with many files, I would prefer to keep the header in each: casual viewers won't know to look in one specific file.

danielinteractive: Ah, I just meant for this `Integration with knitr` piece. That should not be duplicate. Because otherwise it is not clear which file generates the actual `.Rd` file output.

tests/testthat/test-helpers_knitr_NextBest.R
danielinteractive: Ah, ok, then we can keep this as is.

R/helpers_knitr_NextBest.R
```diff
@@ -1,5 +1,3 @@
-# Integration with knitr ----
-#'
 #' @description `r lifecycle::badge("experimental")`
```
danielinteractive, line 1: this part would be important to only include once in one of the R files. otherwise we would get roxygen2 to "overwrite" this multiple times, and for further maintenance it would be unclear which file to edit (or all of them need to be consistently updated, which does not make sense)

tests/testthat/test-Design-methods.R
```diff
@@ -153,6 +154,8 @@ test_that("simulate-TDSamplesDesign produces consistent results", {
   expect_snapshot(result)
 })
 
+# TDDEsign ----
```
danielinteractive, line 157:
```suggestion
# TDDesign ----
```

R/helpers_knitr_NextBest.R
Puzzled-Face: Yes.  For `helpers_knitr_NextBest.R`.  Do you want me to do it for the other `helpers_knitr_xxxx.R` files?

danielinteractive: Yes please check that there is no duplicate documentation, otherwise please remove it and only keep a single instance of the documentation

Puzzled-Face: Done.

danielinteractive (approved):
Thanks @Puzzled-Face 👍🏼

PR #825, opened 2024-06-04
change from roche to openpharma repo

danielinteractive:


Puzzled-Face:
There are some remaining references to `github.com/Roche/crmPack` in the `design` folder.  They are mostly in .html files, which are caught by `.gitignore` files, but there are some in `.Rmd` files, which I believe are not caught.  Otherwise, I think you have everything.

PuzzledFace (commented):
See comment regarding files in `design`.

danielinteractive:
Thanks @Puzzled-Face , design files generally don't need to be updated, so that is ok

PR #830, opened 2024-06-07
829 write a knit print vignette 1

Puzzled-Face:
Also includes a small bug fix relating to the rendition of dose grids.


R/helpers_knitr_GeneralData.R
```diff
@@ -2,24 +2,36 @@
...
   assert_numeric(grid, lower = 0, min.len = 2, unique = TRUE, finite = TRUE, sorted = TRUE, any.missing = FALSE)
   assert_character(units, len = 1)
 
   n <- length(grid)
   units <- h_prepare_units(units)
+  if (is.na(fmt)) {
+    formattedGrid <- as.character(grid)
+  } else {
+    formattedGrid <- sprintf(fmt, grid)
+  }
```
danielinteractive, line 23:
```suggestion
  formattedGrid <- if (is.na(fmt)) {
    as.character(grid)
  } else {
    sprintf(fmt, grid)
  }
```

README.md
```diff
@@ -11,7 +11,7 @@ status](https://www.r-pkg.org/badges/version/crmPack)](https://CRAN.R-project.or
 [![Project Status: Active - The project has reached a stable, usable
 state and is being actively
 developed.](https://www.repostatus.org/badges/latest/active.svg)](https://www.repostatus.org/#active)
-[![License](https://img.shields.io/badge/License-Apache_2.0-blue.svg)](https://opensource.org/license/apache-2-0)
+[![License](https://img.shields.io/badge/License-Apache_2.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
```
danielinteractive, line 14: needs to be regenerated to be consistent with the `Rmd` file regarding this line?

danielinteractive (approved):
Thanks @Puzzled-Face , looks nice

R/helpers_knitr_GeneralData.R
Puzzled-Face: Beyond style, is there any benefit of this suggested change?  I prefer the original as I still think of `if` as a statement rather than a function, but I'll make the change anyway?

danielinteractive: yes, better readability (I always suggest this change in this kind of pattern), yeah `ifelse` can be problematic

PR #831, opened 2024-06-12
First draft of multiple chain discussion document

Puzzled-Face:
First draft of the discussion of implementing multiple MCMC chains.  This focusses on how to adapt the `mcmc` method, not on potential downstream consequences.

As I explain, I favour creation of (a) new `ChainInitialiser` class(es).  As I wrote the document, I realised the current handling of some JAGS options (eg `thin` and `n.iter`) is sub-optimal.  The fix is relatively straightforward.  I will start work on it soon, but would appreciate feedback on the general approach I am proposing as I do so.


design/824-add-nchains-option-to-mcmc/discussion.qmd
```diff
@@ -0,0 +1,1202 @@
...
+  library(crmPack)
+  library(knitr)
+  library(kableExtra)
+  library(mvtnorm)
+  library(checkmate)
+  library(dplyr)
+  library(rjags)
+})
+
+# The Validate class needs to be defined in the Quarto document.  Perhaps because it's not exported from `crmPack`?
```
danielinteractive, line 23: yes, I guess because of that

design/824-add-nchains-option-to-mcmc/discussion.qmd
```diff
@@ -0,0 +1,1202 @@
...
+```{r}
+#| eval: FALSE
+#| label: example-init-slot
+
+init <- function() {
+  list(theta = c(0, 1))
+}
+```
+
+Clearly, what values can be considered extreme depend on the prior distribution of the model. These
```
danielinteractive, line 84:
```suggestion
Clearly, what values can be considered extreme depends on the prior distribution of the model. These
```

design/824-add-nchains-option-to-mcmc/discussion.qmd
```diff
@@ -0,0 +1,1202 @@
...
+  thin = 2
+)
+samples
+```
+
+# Simplistic support for multiple chains
+
+By modifying the
+function`in a`Model`'s`init`slot to take a parameter that defines the number of chains required, multiple`inits\`
+can be easily supported without breaking existing code:
```
danielinteractive, line 142: broken text? or ok?

danielinteractive (commented):
Thanks @Puzzled-Face , very thorough document! Maybe I missed it when browsing through, but can you add a simple example how the new user interface would look like when they want to use multiple chains with default or custom starting values?

design/824-add-nchains-option-to-mcmc/discussion.qmd
Puzzled-Face: If that's my only spelling mistake, I'm amazed!

design/824-add-nchains-option-to-mcmc/discussion.qmd
Puzzled-Face: Broken text.  Now fixed.  If you agree that using a separate `initialise` function is the way to go, then this section (and the following one) _could_ be deleted.  On the other hand, keeping them might provide useful context in the future...

Puzzled-Face:
> Thanks @Puzzled-Face , very thorough document! Maybe I missed it when browsing through, but can you add a simple example how the new user interface would look like when they want to use multiple chains with default or custom starting values?

Is "thorough" a euphemism for "long"? ;-)  It was considerably more work than I expected, but I think the effort was worthwhile.

There's an implicit example of using multiple chains with default values in the section on `mcmc` towards the end of the document [line 996 of the source document]:

```
threeChains <- mcmc(emptyData, myModel, options = LegacyChainOptions(3))
```

I'll add a more direct example immediately before "Other ChainOptions classes"...

danielinteractive (approved):

Issue #815, opened 2024-04-30
`simulate-DADesign` fails


Puzzled-Face:
```
# DADesign ----

test_that("simulate-DADesign produces consistent results", {
  opts <- McmcOptions(
    burnin = 100L,
    step = 2L,
    samples = 200L,
    rng_kind = "Mersenne-Twister",
    rng_seed = 311714
  )
  design <- .DefaultDADesign()

  actual <- simulate(
    design,
    nsim = 2,
    seed = 311714,
    mcmcOptions = opts,
    truthTox = probFunction(design@model, alpha0 = 2, alpha1 = 3),
    truthSurv = function(x, onset = 15) {
      a <- pexp(28, 1 / onset, lower.tail = FALSE)
      1 - (pexp(x, 1 / onset, lower.tail = FALSE) - a) / (1 - a)
    }
  )
  expect_snapshot(actual)
})
```

Fails with the following error:
![image](https://github.com/Roche/crmPack/assets/133956382/e455c3b9-b4cb-456c-81bb-1b788d7cfcb6)

This is therefore a failure in execution, not of expectation.

PR #851, opened 2024-08-02
Allow zero follow-up times in DataDA objects.

### LINKED ISSUE #850: `DataDA` does not allow zero follow-up time
If a `DADesign`'s model is updated on the same day as a participant enrols (so that this participant's follow-up time is zero), the following error is reported:

```
Error in `validObject()`:
! invalid class "DataDA" object: u must be of type double, nObs length, non-negative and not greater than Tmax
```

`u` is a vector of DLT-free survival times.

This seems unnecessarily strict: two participants might well enrol on the same day.  Or what if a participant *immediately* reports a DLT?  Is there a technical reason why survival must be strictly positive?  If not, I feel the `u`s should simply be non-negative rather than strictly positive.

The error message is also inconsistent with the code, the relevant part of which is:

```
  v$check(
    test_numeric(object@u, upper = object@Tmax, len = object@nObs, any.missing = FALSE) &&
      all(object@u > 0),
    "u must be of type double, nObs length, non-negative and not greater than Tmax"
  )
```

#815 _may_ be related.

Puzzled-Face:
Close #850.


Puzzled-Face:
This edit allows the test mentioned in #815 to pass locally.

NEWS.md
```diff
@@ -1,4 +1,4 @@
-# Version 2.0.0.9158
+# Version 2.0.0
```
danielinteractive, line 1: no worries about fixing this now, because the version bump action in GHA will change this again. we can just do it at the end before the CRAN release

danielinteractive (approved):
Thanks @Puzzled-Face , I think it should be ok to allow `u = 0` times

NEWS.md
Puzzled-Face: The version mismatch caused an error in the [version check job](https://github.com/openpharma/crmPack/actions/runs/10214239569/job/28261229637) of the CI.

danielinteractive: OK, then please adjust the `DESCRIPTION` version instead of the `NEWS` version

PR #867, opened 2024-10-03
Fix #866

### LINKED ISSUE #866: `tidy-DataOrdinal` produces incorrect results
Consider

```
.DefaultDataOrdinal() %>% tidy()
# A tibble: 10 × 11
      ID Cohort  Dose Placebo  NObs NGrid DoseGrid   XLevel Cat0  Cat1  Cat2 
   <int>  <int> <dbl> <lgl>   <int> <int> <list>      <int> <lgl> <lgl> <lgl>
 1     1      1    10 FALSE      10    10 <dbl [10]>      1 FALSE FALSE FALSE
 2     2      2    20 FALSE      10    10 <dbl [10]>      2 FALSE FALSE FALSE
 3     3      3    30 FALSE      10    10 <dbl [10]>      3 FALSE FALSE FALSE
 4     4      4    40 FALSE      10    10 <dbl [10]>      4 FALSE FALSE FALSE
 5     5      5    50 FALSE      10    10 <dbl [10]>      5 FALSE FALSE FALSE
 6     6      5    50 FALSE      10    10 <dbl [10]>      5 FALSE TRUE  FALSE
 7     7      5    50 FALSE      10    10 <dbl [10]>      5 FALSE FALSE FALSE
 8     8      6    60 FALSE      10    10 <dbl [10]>      6 FALSE FALSE FALSE
 9     9      6    60 FALSE      10    10 <dbl [10]>      6 FALSE TRUE  FALSE
10    10      6    60 FALSE      10    10 <dbl [10]>      6 FALSE FALSE TRUE
```

By definition, `Cat0` should be equal to `!(Cat1 || Cat2)`, but it is `FALSE` for all rows.  The fix is straightforward: add an additional `mutate` to correct the default handling of `Cat0`.

Puzzled-Face:


tests/testthat/test-Data-methods.R
```diff
@@ -656,3 +656,17 @@ test_that("tidy-DataGeneral creates the correct tibble", {
   expected$Tox <- c(FALSE, TRUE, FALSE)
   expect_equal(tidy(d), expected)
 })
+
+# See https://github.com/openpharma/crmPack/issues/866
```
danielinteractive, line 660:
```suggestion
```
we don't do that, note that via GitHub blame we can find out later where which test was added/edited

danielinteractive (commented):
thanks @Puzzled-Face

danielinteractive (approved):

Issue #870, opened 2025-01-14
More informative error message when misspelling a parameter in a call to `Data()`


PuzzledFace-Astellas:
I carelessly wrote

```
observedData <- Data(
  x = c(1, 1, 1, 2, 2, 2),
  y = c(0, 0, 0, 0, 1, 0),
  dataGrid = c(1, 2, 4, 8, 16, 30, 50, 100),
  ID = 1L:6L,
  cohort = rep(1:2, each = 3)
)
```

and was told

```
Error in Data(x = c(1, 1, 1, 2, 2, 2), y = c(0, 0, 0, 0, 1, 0), dataGrid = c(1,  : 
  Assertion on 'x' failed: Must be a subset of the empty set, i.e. also empty.
```

It took me ages to realise that I'd typed `dataGrid` rather than `doseGrid`.

The assertion on `doseGrid` at the start of `Data()` is currently:

```
assert_numeric(doseGrid, any.missing = FALSE, unique = TRUE)
```

Changing this to

```
assert_numeric(doseGrid, any.missing = FALSE, unique = TRUE, min.len = 1)
```
at least gives an error message that references `doseGrid`:

```
Error in Data(x = c(1, 1, 1, 2, 2, 2), y = c(0, 0, 0, 0, 1, 0), dataGrid = c(1,  : 
  Assertion on 'doseGrid' failed: Must have length >= 1, but has length 0.
```

I can't think of a reason why we would want to permit empty `doseGrid`s.

PR #871, opened 2025-01-14
Completes #870

PuzzledFace-Astellas:
I made the test for `min.len` conditional on having a non-NULL `x`.  This allows other existing tests on `Data` to pass without modification.


danielinteractive (approved):
Thanks @PuzzledFace-Astellas !

PR #883, opened 2025-10-21
skip vignette builds during check for faster build

danielinteractive:
based on recent update in https://github.com/insightsengineering/r.pkg.template/pull/286


danielinteractive:
Hi @cicdguy , sorry quick question, maybe I am using your new feature wrongly - in https://github.com/openpharma/crmPack/actions/runs/18685122278/job/53275581310?pr=883 I am getting 
an error which I can't reproduce locally, do you have an idea?

